PR #11, opened 2017-11-03
Adding documentation for providers wrapped in functions

bmmpxf:
See https://github.com/trufflesuite/truffle/issues/348

### DIFF
--- a/public/docs/advanced/configuration.md
+++ b/public/docs/advanced/configuration.md
@@ -79,6 +79,46 @@ For each network, if unspecified, transaction options will default to the follow
 * `from`: From address used during migrations. Defaults to the first available account provided by your Ethereum client.
 * `provider`: Default web3 provider using `host` and `port` options: `new Web3.providers.HttpProvider("http://<host>:<port>")`
 
+#### Accessing only one of multiple networks
+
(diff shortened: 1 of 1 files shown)


public/docs/advanced/configuration.md
```diff
@@ -79,6 +79,46 @@ For each network, if unspecified, transaction options will default to the follow
 * `from`: From address used during migrations. Defaults to the first available account provided by your Ethereum client.
 * `provider`: Default web3 provider using `host` and `port` options: `new Web3.providers.HttpProvider("http://<host>:<port>")`
 
+#### Accessing only one of multiple networks
```
gnidan: Note that this section only applies when using `provider` - using `host`/`port` instead means you don't have to do this function wrapping. Not sure if you want to mention that or not. Otherwise looks great!

bmmpxf:
@gnidan Okay, I pushed a new commit that clarifies more of the reasons why to use `provider` vs `host` and `port`. I also changed the example to be more real-world useful (as since `host` and `port` don't suffer the function wrapping issue, there didn't seem to be a point to an example with HttpProvider, so I changed the provider to HDWalletProvider).

Please take a look and let me know if I've got the argumentation correct. Thanks!

gnidan:
Looks great!

PR #16, opened 2017-11-08
Updating client docs to include Truffle Develop

bmmpxf:


public/docs/getting_started/client.md
```diff
@@ -1,20 +1,54 @@
...
 There are many Ethereum clients to choose from. We recommend using different clients when developing and deploying.
 
-# When Developing
+## Developing locally
+
+### Truffle Develop
+
+We recommend using Truffle Develop, an Ethereum client built directly into Truffle which contains an in-memory blockchain that runs only on your development machine.
+
+It processes transactions instantly instead of waiting for the default block time, so you can test that your code works quickly, and it tells you immediately when your smart contracts run into errors. It also makes a great client for automated testing.
```
tcoulter: New paragraph; should probably use "Truffle Develop" instead of It"

public/docs/getting_started/client.md
```diff
@@ -1,20 +1,54 @@
...
+
+Because Truffle Develop is built into Truffle, no external installation is required. To run Truffle Develop, type the following in a terminal:
+
+```shell
+truffle develop
+```
+
+This will run the client on `http://localhost:9545`. it will display the first 10 accounts, and a deterministic mnemonic for the blockchain. Truffle Develop uses the same mnemonic upon every launch:
```
tcoulter: Needs capital `I` in "it" at beginning of second sentence.

public/docs/getting_started/client.md
```diff
@@ -1,20 +1,54 @@
...
+
+```
+candy maple cake sugar pudding cream honey rich smooth crumble sweet treat
+```
+
+Once the Truffle Develop console is launched, you can run all of the commands you are used to, but omitting the `truffle` prefix. So, for example, to compile your smart contract, instead of typing `truffle compile` on the terminal, just type `compile` in the Truffle Develop console.
```
tcoulter: "Truffle Develop console" is being introduced here as a new thing. (i.e., to a new user, it might create the question, is there a difference between "Truffle Develop" and "Truffle Develop console"?). Additionally, "all the commands you're used to" doesn't mean anything to first time readers. I'd rephrase like this:

> Once launched, Truffle Develop will provide you with a console you can use to run all available Truffle commands. These commands are input by omitting the `truffle` prefix. So, for example, to compile your smart contracts, instead of typing `truffle compile`, you need to only type `compile`.

public/docs/getting_started/client.md
tcoulter: > a deterministic mnemonic

"Deterministic mnemonic" feels like too much detail here. I'd phrase it as the following: 

> It will display the first 10 accounts and the mnemonic used to create those accounts (for more information on account mnemonics, please see link xyz). Truffle Develop uses the same mnemonic every time to make developing your applications as easy as possible: "

I can help find a good link about mnemonics if you can't find one.

public/docs/getting_started/client.md
tcoulter: I see you tried to pare down what was written for the TestRPC, but this effectively removed a lot of "the sell". i.e., we need to convince people to use it, and if you compare the new description of the TestRPC to this description of Truffle Develop, the TestRPC is sold a lot better. We should take more from the TestTPC's "sell" as we want to encourage people to use Truffle Develop instead. Here's my attempt at merging the two. Note that "in memory blockchain" is no longer correct, so I removed that.

> We recommend using Truffle Develop, a development blockchain built directly into Truffle. It processes transactions instantly instead of waiting for the default block time -- so you can test that your code works quickly -- and it tells you immediately when your smart contracts run into errors. It also makes a great client for automated testing, and Truffle knows how to use its special features to speed up test runtime by almost 90% compared to other clients. 

See my notes below about how to deemphasize the TestRPC.

public/docs/getting_started/client.md
```diff
@@ -1,20 +1,54 @@
...
-# When Developing
+## Developing locally
+
+### Truffle Develop
+
+We recommend using Truffle Develop, an Ethereum client built directly into Truffle which contains an in-memory blockchain that runs only on your development machine.
+
+It processes transactions instantly instead of waiting for the default block time, so you can test that your code works quickly, and it tells you immediately when your smart contracts run into errors. It also makes a great client for automated testing.
+
+Because Truffle Develop is built into Truffle, no external installation is required. To run Truffle Develop, type the following in a terminal:
```
tcoulter: I prefer "into" instead of "in" in the phrase "in a terminal here". Either works however if you have strong opinions.

public/docs/getting_started/client.md
```diff
@@ -1,20 +1,54 @@
+# Choosing an Ethereum client
+
 There are many Ethereum clients to choose from. We recommend using different clients when developing and deploying.
 
-# When Developing
+## Developing locally
```
tcoulter: I don't think we should highlight "locally" here. We want to paint `truffle develop` (and eventually Ganache) as the what you use to develop, full stop. So "When Developing" fits better, because "Developing locally" suggests you should use something else when developing on a blockchain that exists on a different computer.

public/docs/getting_started/client.md
```diff
@@ -1,20 +1,54 @@
...
+You can also use [EthereumJS TestRPC](https://github.com/ethereumjs/testrpc). Like Truffle Develop, it is an in-memory blockchain that runs only on your development machine. Truffle knows how to use its special features to speed up test runtime by almost 90%.
+
+To run the TestRPC, open a new terminal and type:
+
+```shell
+testrpc
+```
 
-* EthereumJS TestRPC: [https://github.com/ethereumjs/testrpc](https://github.com/ethereumjs/testrpc)
+The TestRPC will start with a different seed each time (and so a different mnemonic and account list), but you can run the TestRPC with a specific mnemonic in order to regenerate the same accounts:
```
tcoulter: As mentioned above, this is a pretty good sell. Also, providing an example of how to use the TestRPC here emphasizes the TestRPC instead of deemphasizing it. We should like pare it down to something like this:

> You can also use the [EthereumJS TestRPC](https://github.com/ethereumjs/testrpc). It's a separate command-line blockchain you can use for both development and automated testing, and works well with Truffle. Check out it's README and documentation for more information.

Note that I *did* emphasize that it's separate, hopefully helping to sell `truffle develop` as being integrated, since users are really liking not having to download and install another application.

public/docs/getting_started/client.md
bmmpxf: Oh, I think I'm confused about the scope of Truffle Develop then. I thought it was just a local program you ran when developing. Is that not the case? Or am I confusing the application with the network here?

public/docs/getting_started/client.md
bmmpxf: Thanks for the rewrite, that's very helpful.

Can you provide clarification about the "in-memory blockchain" not being correct? I took that right from @gnidan 's oral description, and I've been using that phrase pretty much everywhere (so if it's not correct, I'll want to change it in a bunch of places).

public/docs/getting_started/client.md
bmmpxf: Using "into" is fine there. If it were a preceding clause, I'd prefer "in" (as in, "In a terminal, type the following:").

public/docs/getting_started/client.md
bmmpxf: If you have something authoritative on mnemonics, that would be appreciated!

public/docs/getting_started/client.md
bmmpxf: Just to make sure, I know Truffle Develop is the application, but is it correct to refer to the thing we types commands at as the "Truffle Develop console"? I'd like to be able differentiate the thing from the application as a whole.

public/docs/getting_started/client.md
bmmpxf: Thanks for this. I wasn't very clear on how we wanted to sell TestRPC, or how we wanted to differentiate it from Truffle Develop.

bmmpxf:
@tcoulter Thanks for the detailed review! Some questions back to you before I fix it up.

public/docs/getting_started/client.md
tcoulter: As of TestRPC 4.0, the blockchain is no longer stored in memory, as the blockchain takes up a lot of data, which effectively makes it a built in memory leak (not good). Instead, it's now stored on the filesystem and has a database attached to it. For now, each launch of the TestRPC creates a temp directory to store that data to maintain the same behavior it hand when it was initially created.

public/docs/getting_started/client.md
tcoulter: That's sounds good to me, in general. Here, the console hasn't yet been introduced (i.e., a new user may confuse the two) so I wanted to ensure people understand that `truffle develop` creates a console.

Additionally (new thought) we should link to [this page](https://github.com/trufflesuite/trufflesuite.com/blob/master/public/docs/getting_started/console.md) on the site in (or near) this paragraph so people can read more on how to use the console.

public/docs/getting_started/client.md
tcoulter: It is local to the computer it runs on, in general, but doesn't necessarily have to be. For instance, some people like to start up an Azure instance that exists somewhere else on the cloud to use Truffle.

From a branding perspective, we want to solidify Truffle Develop as what you use for development. Highlighting "locally" suggests that you'd use something different were you not wanting to develop locally, but that's such a rare case and generally not recommended unless needed.

bmmpxf:
Thanks for all the clarifications. Second pass is committed.

public/docs/getting_started/client.md
```diff
@@ -1,20 +1,54 @@
...
+truffle develop
+```
+
+This will run the client on `http://localhost:9545`. It will display the first 10 accounts and the mnemonic used to create those accounts. ([Read more about account mnemonics](https://github.com/bitcoin/bips/blob/master/bip-0039.mediawiki).) Truffle Develop uses the same mnemonic every time to make developing your applications as easy as possible:
+
+```
+candy maple cake sugar pudding cream honey rich smooth crumble sweet treat
+```
+
+Once launched, Truffle Develop will provide you with a console you can use to run all available Truffle commands. These commands are input by omitting the `truffle` prefix. So, for example, to compile your smart contracts, instead of typing `truffle compile`, you need to only type `compile`.
```
tcoulter: Let's add this line after this paragraph:

> To read more about interacting with the console, see the [Using the Console](/docs/getting_started/console) section.

public/docs/getting_started/client.md
```diff
@@ -1,20 +1,54 @@
...
+```shell
+testrpc
+```
+
+The TestRPC will start with a different seed each time (and so a different mnemonic and account list), but you can run the TestRPC with a specific mnemonic in order to regenerate the same accounts:
+
+```shell
+testrpc --mnemonic word1 word2 word3 ...
+```
+
```
tcoulter: We should take out lines 28 through 39 here as we want to deemphasize its usage.

public/docs/getting_started/client.md
```diff
@@ -1,20 +1,54 @@
...
 
 * Geth (go-ethereum): [https://github.com/ethereum/go-ethereum](https://github.com/ethereum/go-ethereum)
 * WebThree (cpp-ethereum): [https://github.com/ethereum/cpp-ethereum](https://github.com/ethereum/cpp-ethereum)
 * Parity: [https://github.com/paritytech/parity](https://github.com/paritytech/parity)
 * More: [https://www.ethereum.org/cli](https://www.ethereum.org/cli)
 
-There are many official and unofficial Ethereum clients available for you to use. You should use these clients after you've sufficiently tested your dapp with the EthereumJS TestRPC and you're ready to deploy to your desired Ethereum network. These are full client implementations that include mining, networking, blocks and transaction processing, and Truffle can deploy to these clients without any extra configuration.
+These are full client implementations that include mining, networking, blocks and transaction processing. You should use these clients after you've sufficiently tested your dapp with Truffle Develop or the TestRPC and you're ready to deploy to your desired Ethereum network. Truffle can deploy to these clients without any extra configuration.
```
tcoulter: Like this separation. 👏

PR #19, opened 2017-11-09
Updating "Compiling contracts" page for Truffle 4

bmmpxf:
Includes references to Truffle Develop, and also the new bare bones `truffle init` project.


public/docs/getting_started/compile.md
```diff
@@ -1,43 +1,59 @@
-# Location
+# Compiling contracts
 
-All of your contracts are located in the `./contracts` directory. By default you're given an example Solidity [contract](http://solidity.readthedocs.org/en/latest/contracts.html) file and an example Solidity [library](http://solidity.readthedocs.org/en/latest/contracts.html#libraries) file, both with filenames ending in `.sol`. Although a Solidity library is different than a contract, for documentation purposes we're going to lump these into the same word, "contracts".
+## Location
 
-# Command
+All of your contracts are located in your project's `contracts/` directory. As contracts are written in [Solidity](LINK), all files containing contracts will have a file extension of `.sol`. Associated Solidity [libraries](http://solidity.readthedocs.org/en/latest/contracts.html#libraries) will also have a `.sol` extension.
```
tcoulter: Whoops. Looks like there's a LINK here that's not fleshed out. Probably this link is best? https://solidity.readthedocs.io/en/develop/

public/docs/getting_started/compile.md
```diff
@@ -1,43 +1,59 @@
...
 
-To compile your contracts, simply run:
+With a bare Truffle [project](./project) (created through `truffle init`), you're given a single `Migrations.sol` file that is responsible for deploying your contracts to the blockchain. If you're using a [Truffle Box](/boxes), you will have multiple files here.
```
tcoulter: This isn't quite true about the Migrations contract. The Migrations contract itself simply _records_ successful migrations; it doesn't do the deploying. The migrations feature in Truffle does the deploying, and that features uses the Migrations contract. 

This should probably be changed to something like: 
> you're given a single `Migrations.sol` file that helps in the deployment process.

public/docs/getting_started/compile.md
```diff
@@ -1,43 +1,59 @@
...
-$ truffle compile
+## Command
+
+If you are using Truffle Develop, type the following into the Truffle Develop console:
+
+```shell
+compile
+```
+
+Otherwise, in a terminal, type the following:
```
tcoulter: Artistic change: the previous sentence had `type the following` in it, and this sentence ends that way. I'd switch it up to something like `in a terminal, type this instead`. Up to you.

tcoulter: On a separate note, I don't actually think we need to call out commands specific to `truffle develop`. e.g., if we do that here, then we have to do that everywhere within our documentation. The `truffle develop` and `console` documentation has enough info for using those consoles. So on second thought, I'd remove the `truffle develop` example here.

public/docs/getting_started/compile.md
```diff
@@ -1,43 +1,59 @@
...
+### Importing dependencies via file name
 
-To import contracts from a separate file, simply write the following command, where `AnotherContract.sol` is relative to the current contract being written. This will make all contracts within `AnotherContract.sol` available to the current source file.
+To import contracts from a separate file, simply use the following command:
 
 ```
 import "./AnotherContract.sol";
 ```
 
+This will make all contracts within `AnotherContract.sol` available to the current source file. Here, `AnotherContract.sol` is relative to the path of the current contract being written.
```
tcoulter: Aside: If you take my suggestion above word for word, we're using `source file` twice. I'd change one or the other in that case.

public/docs/getting_started/compile.md
bmmpxf: Hmm, I thought we wanted to encourage people to use the Truffle Develop console over the terminal. I was thinking that if anything we would take out the terminal instructions here.

Granted, yes that means that eventually we might want to change a bunch of commands in our docs, but whatever we write is implicitly a recommendation, so if we're not using Truffle Develop, then others might not either. Or is Truffle Develop for a specific audience?

tcoulter: We do want to encourage people, but if we follow this structure every example on every page would need its `truffle develop` equivalent, which is a bit distracting (i.e., just remove `truffle` from the command). `Choosing and Ethereum Client` is a good place for pumping up `truffle develop`, and tutorials are as well. If you do want to mention `truffle develop` on this page, I'd recommend using a sentence similar to `This command is also available in Truffle Develop` and linking `Truffle Develop` to the page that describes it.

tcoulter:
This one looks good to go, assuming you made the changes with the 👍 's.

bmmpxf:
Great, thanks! Yes, 👍 implies they are done (or will be).

PR #23, opened 2017-11-10
Update getting started

ogryzek:
Include truffle init in installation

### DIFF
--- a/public/docs/getting_started/installation.md
+++ b/public/docs/getting_started/installation.md
@@ -4,6 +4,12 @@
 $ npm install -g truffle
 ```
 
+Initialize your working directory (run `truffle help` for a list of commands)
+
+```
+$ truffle init
+```
+
 ## Requirements
 
 * NodeJS 5.0+ recommended.


ogryzek:
I see `truffle init` is mentioned in [Creating a Project](http://truffleframework.com/docs/getting_started/project).

I'm not sure my commit is a _better_ solution, but I found it a little confusing that `truffle develop` didn't work after following [installation](http://truffleframework.com/docs/getting_started/installation), and seeing it in the [next section](http://truffleframework.com/docs/getting_started/client).

bmmpxf:
Hi @ogryzek . Thanks for your suggestion; I see your point. I think it might be clearer if we just switched the order of the sections:

1. Installation
2. Creating a project *(was 3.)*
3. Choosing an Ethereum client *(was 2.)*

I think that might alleviate confusion, and would make sure that people had a project before attempting to run `truffle develop` (assuming they were reading in order). Would that achieve the same goal as your PR?

ogryzek:
@bmmpxf I think that would be perfect!

bmmpxf:
You got it! Done with https://github.com/trufflesuite/trufflesuite.com/commit/ce907472925351a9310ef88d1caaf4ce066bf93e.

PR #24, opened 2017-11-10
Update debugging-a-smart-contract.md

cashlionjp:
truffle init no longer accepts a project template name as an arg in v4+


bmmpxf:
Right you are @cashlionjp . We're still in the process of updating the docs for Truffle 4. If you see anything else that needs updating, please let us know. Thanks!

PR #33, opened 2017-11-22
Adding basic docs about Ganache

bmmpxf:


public/docs/ganache/using.md
```diff
@@ -0,0 +1,64 @@
+# Working with Ganache
+
+[Ganache](/ganache) is an application that generates a personal blockchain you can use to run tests, execute commands, deploy contracts, and inspect state.
```
tcoulter: "an application that generates" - stronger if just "is". 

i.e., 

> [Ganache](/ganache) is a personal blockchain for Ethereum development you can use to deploy contracts, develop your applications, and run tests. 

I put "run tests" last as that's less of the focus of Ganache, although possible.

public/docs/ganache/using.md
```diff
@@ -0,0 +1,64 @@
...
+
+![Ganache logo](/docs/ganache/images/ganache-logo.png)
+
+It is available in both a graphical format and as a terminal application (which was previously known as the TestRPC). Ganache is available for Windows, macOS, and Linux.
```
tcoulter: I'd call these a "desktop application" and "command line tool"

> It is available as both a desktop application as well as a command line tool (the command line tool was formerly known as the TestRPC).

tcoulter: Also, `macOS` is the less common way to refer to a Mac, although probably technically correct. Can we revert the more colloquial "Mac"?

public/docs/ganache/using.md
```diff
@@ -0,0 +1,64 @@
...
+
+## Installation
+
+### Ganache
+
+Installation of Ganache is only a few steps. [Download](https://github.com/trufflesuite/ganache/releases) the appropriate version for your OS:
```
tcoulter: When I read this as a "few steps", it actually sounded harder than it is. I'd almost remove the step count and just call it "easy": i.e., 

> Installing Ganache is easy. [Download](https://github.com/trufflesuite/ganache/releases) the appropriate version for your OS then double click on the file:

public/docs/ganache/using.md
```diff
@@ -0,0 +1,64 @@
...
+
+* Windows: `Ganache-a.b.c.appx`
```
tcoulter: Nit picky, but can we make these `Ganache-*.appx`, etc? This way we don't have to change the docs if we choose to change the release format.

public/docs/ganache/using.md
```diff
@@ -0,0 +1,64 @@
...
+* macOS: `Ganache-a.b.c.dmg`
+* Linux: `Ganache-a.b.c.AppImage`
+
+Double click on the downloaded file, follow the prompts, and you're up and running.
```
tcoulter: Given the changes above, may not need this.

public/docs/ganache/using.md
```diff
@@ -0,0 +1,64 @@
...
+
+### Ganache CLI
+
+If you are interested in only the command-line application (previously known as the TestRPC), you can get it through npm.
```
tcoulter: I'd remove "only" as you can be interested in both.

tcoulter: Maybe switch "application" with "version" /artistic

public/docs/ganache/using.md
```diff
@@ -0,0 +1,64 @@
...
+
+```shell
+npm install -g ganache-cli
+```
+
+This page will focus only on the graphical interface. Please see the [README](https://github.com/trufflesuite/ganache-cli/blob/master/README.md) for more information on Ganache CLI.
+
+## Main interface
+
+When you first launch Ganache, the screen will show some details about the server, and also list out a number of accounts. You'll note that each account has been given 100 ether, to make testing of transactions much easier.
```
tcoulter: This is actually not true. On your _very first_ run, you're asked whether or not you want to allow Google analytics tracking. Should we add that? 

Switch "testing of transactions" with "developing your application" easier (i.e., focus on development in general instead of testing specifically)

public/docs/ganache/using.md
```diff
@@ -0,0 +1,64 @@
...
+
+![Ganache](/docs/ganache/images/accounts.png)
+
+*Ganache accounts*
+
+There are four pages available:
+
+* The **Accounts** page shows the accounts generated. This is the default view.
```
tcoulter: I'd add "and their balances", to highlight that balances change as you make transactions

public/docs/ganache/using.md
bmmpxf: I'll do a slight rephrase of the parenthetical to remove duplication, otherwise 👍 .

public/docs/ganache/using.md
bmmpxf: Fair point, though I try to avoid using the words "easy", "simple" etc. to describe an application's functions.

I'll just take that whole sentence out, as they'll see below just how little there is in the section!

Also, I would want to keep the double-click below the file name, as that more directly reflects the order of operations. Also keeps the section from ending on a bulleted list.

public/docs/ganache/using.md
bmmpxf: Oh good point. I'll add a note about Google Analytics.

bmmpxf:
@tcoulter Thanks for the review!

PR #34, opened 2017-11-24
Better description on how to get started.

gcurelet:
Explained Metacoin get started process.


bmmpxf:
Hi @gcurelet . Thanks for the suggestion! I like the way you put back in the MetaCoin example. We definitely want to make sure that people know that project is still there.

If you have any other ideas or suggestions, please let us know. Thanks!

PR #40, opened 2017-12-04
Update pet-shop.md

joesan:
I just blindly followed the pet shop tutorial and ran into errors:

```
Could not connect to your Ethereum client. Please check that your Ethereum client:
    - is running
    - is accepting RPC connections (i.e., "--rpc" option is used in geth)
    - is accessible over the network
    - is properly configured in your Truffle configuration file (truffle.js)
```

It might be a good idea to have some hints on atleast running a test ethereum client. 

Does this make sense?

### DIFF
--- a/public/tutorials/pet-shop.md
+++ b/public/tutorials/pet-shop.md
@@ -35,6 +35,7 @@ There are a few technical requirements before we start. Please install the follo
 
 *   [Node.js v6+ LTS and npm (comes with Node)](https://nodejs.org/en/)
 *   [Git](https://git-scm.com/)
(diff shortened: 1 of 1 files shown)


bmmpxf:
Hi @joesan . We don't need to mention a client as a prerequisite in the tutorial because we specifically invoke `truffle develop` which is built-in to Truffle. This eliminates the need for a separate client.

Now with [Ganache](http://truffleframework.com/ganache) released, we're going to update our tutorial to use that over Truffle Develop (though both will still work).

If I've misunderstood your suggestion, please feel free to reopen. Thanks!

Issue #18, opened 2017-11-09
Use 127.0.0.1 instead of localhost in docs


bmmpxf:
As per https://consensys.slack.com/messages/C1D2T7NLC/

When offline, TestRPC / Truffle Develop fails with an `Invalid JSON RPC response` error.

The problem is that `localhost` needs to be resolved. Switching the network connection string to use `127.0.0.1` instead of `localhost` removes the need for a network lookup.

We need to update our content to reflect this.

bmmpxf:
Addressed with #41 .

PR #41, opened 2017-12-04
Replacing "localhost" with "127.0.0.1" to allow for offline deployment

bmmpxf:
See https://github.com/trufflesuite/trufflesuite.com/issues/18

I know this is a simple search and replace, but I'd like a quick review. Specifically:

* Do we really want to replace every instance of `localhost` with `127.0.0.1`? Any exceptions? What about the apps that currently say they run on `localhost:8080`?
* Is it going to be obvious to everyone that 127.0.0.1 is local? Instead of saying "It runs on 127.0.0.1:8545", should we say "It runs on the local 127.0.0.1:8545"?
* I replaced all docs and tutorials, with the exception of the one that upgrades from 2 to 3, as it seemed like it was either not relevant or too old to matter.


public/docs/advanced/truffle-with-metamask.md
```diff
@@ -88,12 +88,12 @@ Click the account icon in the upper-right to create new accounts, the first 10 o
 
 ## Using MetaMask with Truffle Develop
 
-Truffle Develop is a command-line application that runs a temporary blockchain that is also used for testing purposes. It runs on `localhost:9545`.
+Truffle Develop is a command-line application that runs a temporary blockchain that is also used for testing purposes. It runs on `127.0.0.1:9545`.
```
benjamincburns: Technically this change _is_ more correct, but I think it's more canonical & reads better to leave this one as-is

public/docs/advanced/truffle-with-metamask.md
```diff
@@ -88,12 +88,12 @@ Click the account icon in the upper-right to create new accounts, the first 10 o
...
 
 Using MetaMask with Truffle Develop is very similar to that of Ganache. There are only a few notable differences:
 
 * The mnemonic is always `candy maple cake sugar pudding cream honey rich smooth crumble sweet treat` and can't be changed.
-* Truffle Develop runs by default on `localhost:9545`, so you'll want to edit the above web3 code to say:
+* Truffle Develop runs by default on `127.0.0.1:9545`, so you'll want to edit the above web3 code to say:
```
benjamincburns: Technically this change is more correct, but I think it's more canonical & reads better to leave this one as-is

benjamincburns: Actually, no - scratch that - don't want to pass the ambiguity on... If anything, we might want to clarify the importance of using `127.0.0.1` instead of `localhost` here...

public/docs/advanced/truffle-with-metamask.md
benjamincburns: Actually, no - scratch that - don't want to pass the ambiguity on... If anything, we might want to clarify the importance of using `127.0.0.1` instead of `localhost` here...

public/docs/getting_started/client.md
```diff
@@ -8,7 +8,7 @@ There are many Ethereum clients to choose from. We recommend different clients d
 
 We recommend [Ganache](/ganache), a personal blockchain for Ethereum development that runs on your desktop. Part of the Truffle Suite, Ganache simplifies dapp development by placing your contracts and transactions front and center. Using Ganache you can quickly see how your application affects the blockchain, and introspect details like your accounts, balances, contract creations and gas costs. You can also fine tune Ganache's advanced mining controls to better suit your needs. Ganache is available for Windows, Mac and Linux, and you can [download it here](/ganache).
 
-Ganache, when launched runs on `http://localhost:7545`. It will display the first 10 accounts and the mnemonic used to create those accounts. ([Read more about account mnemonics](https://github.com/bitcoin/bips/blob/master/bip-0039.mediawiki).)
+Ganache, when launched runs on `http://127.0.0.1:7545`. It will display the first 10 accounts and the mnemonic used to create those accounts. ([Read more about account mnemonics](https://github.com/bitcoin/bips/blob/master/bip-0039.mediawiki).)
```
benjamincburns: Might also be good to add clarity on why `127.0.0.1` and not `localhost` here, too...

Same goes for anywhere else that we're potentially telling the user the URL for the first time.

public/tutorials/bundling-with-webpack.md
```diff
@@ -53,7 +53,7 @@ Since you're bundling the frontend yourself and not using Truffle's default buil
 
 ### Detecting Web3
 
-In the past, Truffle bootstrapped applications to look for an open Ethereum node running at `http://localhost:8545`, but this has proven to be insecure for the user as well as highly unlikely in a real world scenario. Instead, users are most likely to use Metamask or Mist to interact with web applications on the Ethereum network, and you should build your application to support that. Both Metamask and Mist inject a `web3` object when the page is loading so you can hook into their transaction signing processes. Detecting this `web3` object easy:
+In the past, Truffle bootstrapped applications to look for an open Ethereum node running at `http://127.0.0.1:8545`, but this has proven to be insecure for the user as well as highly unlikely in a real world scenario. Instead, users are most likely to use Metamask or Mist to interact with web applications on the Ethereum network, and you should build your application to support that. Both Metamask and Mist inject a `web3` object when the page is loading so you can hook into their transaction signing processes. Detecting this `web3` object easy:
```
benjamincburns: Maybe revert this one, as it literally did use `localhost` in the past.

bmmpxf:
@benjamincburns Changes made. Thanks for your review! If you like the notes I added about why `127.0.0.1` versus `localhost`, I'll go ahead and merge.

public/docs/getting_started/client.md
tcoulter: I think the the note (see below) is odd here. It's almost like we're introducing a topic and warning them about everything that can go wrong before they get a grasp of it. A newcomer wouldn't even know to swap `127.0.0.1` with `localhost`, and instead will input the URL in the address bar as they're told. 

If you think we definitely should keep it, can we pare it down? See example I gave in a comment above.

public/docs/advanced/truffle-with-metamask.md
tcoulter: You have the note (see below) within this file already. I disagree with Ben and think this note might be TMI (in both instances. not just this one). We're displaying a big warning meant to prevent a very minor problem. And if people choose to follow our directions rather than go off the rails, they'll be fine without the warning. 

If we must have the warning, can we pare it down? e.g., 

> **Note**: We recommend specifying `127.0.0.1` instead of `localhost` as it does not require a network connection. Specifying `localhost` when offline will cause connection errors.

public/tutorials/building-dapps-for-quorum-private-enterprise-blockchains.md
```diff
@@ -276,7 +293,32 @@ We originally configured Truffle to point our development environment to the fir
...
+      host: "127.0.0.1",
+      port: 22006,
+      network_id: "*" // Match any network id
+    }
+  }
+};
+```
+=======
    Note that Truffle's contract abstractions use Promises to interact with Ethereum. This can be a little cumbersome on the console as it requires a few extra key strokes to get things done, but within your application it makes control flow a lot smoother. Additionally, take a look at the output we received: We got `42` back, but as an object. This is because Ethereum can represent larger numbers than those natively represented by JavaScript, and so we need an abstraction in order to interact with them.
+>>>>>>> e7d2b4e08b42d85125d36c4519cd67e045d401fe
```
tcoulter: Same thing here. Looks like there's another merge error.

public/docs/advanced/truffle-with-metamask.md
bmmpxf: I left it in, but pared it down even further:

> We recommend specifying `127.0.0.1` instead of `localhost` because the address does not require a network connection and so is more suitable for development.

public/docs/getting_started/client.md
bmmpxf: Removed from here...

public/tutorials/building-dapps-for-quorum-private-enterprise-blockchains.md
bmmpxf: Whoops! Good catch.

bmmpxf:
Thanks for all the feedback @benjamincburns  and @tcoulter ! I think it's much better now.

PR #48, opened 2017-12-18
Update robust-smart-contracts-with-openzeppelin.md

martriay:
Type declaration for `decimals` property should be `uint8` instead of `uint` according to [spec](https://github.com/frozeman/EIPs/blob/94bc4311e889c2c58c561c074be1483f48ac9374/EIPS/eip-20-token-standard.md#decimals).


bmmpxf:
Thanks @martriay ! The spec doesn't lie. :)

PR #51, opened 2017-12-20
Update pet-shop.md

robbieprevost:
Not sure if it's the right place, but I added App.initContract() to the initWeb3 function because initContract is never called. Without initContract being called, the Adopt buttons don't work and MetaMask doesn't pop up for approval of the adoption transaction. There may be a better way to go about calling initContract, but I figured I'd submit this to highlight the issue.

### DIFF
--- a/public/tutorials/pet-shop.md
+++ b/public/tutorials/pet-shop.md
@@ -416,6 +416,7 @@ The front-end doesn't use a build system (webpack, grunt, etc.) to be as easy as
      App.web3Provider = new Web3.providers.HttpProvider('http://localhost:7545');
    }
    web3 = new Web3(App.web3Provider);
+   App.initContract();
    ```
 
 Things to notice:


bmmpxf:
Hi @robbieprevost . I just looked, and `initContract` is being called, as part of the return value of the function `initWeb3`:

https://github.com/truffle-box/pet-shop-box/blob/master/src/js/app.js#L31

I admit it's a little hard to follow, with functions returning outputs from other functions (which in turn call other functions) but it's all there.

As for why the Adopt buttons are disabled, it's hard to tell without seeing the finished tutorial files, but I have tested and updated this tutorials recently, and was able to go through it successfully, so I'm hoping it's merely a typo somewhere.

I'm going to close this PR, but if you think this has been closed in error, please let me know and I'll look into it further. Thanks!

PR #52, opened 2017-12-22
Update Pet-Shop.md

thebkr7:
Removed the "view" word from line 130. With the word view in there, you can never click the adopt button and MetaMask will never pop up.


bmmpxf:
Thanks @thebkr7 ! Looks like we caught this as part of PR #50, just merged in.

0pay:
I Removed the "view" word from line 130 but it still doesn't work when i click the adopt button

PR #55, opened 2017-12-26
Add Testing ether transactions exemple

vincentlg:
I propose this because I had trouble finding the syntax to do that (There may be a simpler way to indicate this...)

### DIFF
--- a/public/docs/getting_started/solidity-tests.md
+++ b/public/docs/getting_started/solidity-tests.md
@@ -145,3 +145,41 @@ contract TestContract {
 ```
 
 Note that Truffle sends Ether to your test contract in a way that does **not** execute a fallback function, so you can still use the fallback function within your Solidity tests for advanced test cases.
+
+Below is another example with ether transactions and parameters. Imagine a crowdfunding campaign contract on which contributors could attach a message to their contribution. Here is an extract of the contract to be tested:
+
+```javascript
+contract Campaign {
+
+    // code of campaign smart contract 
+    ... 
+    
+    function contribute(string _message) public payable {
+        amountRaised += msg.value;
+        // todo store contribution _message in the contributions array
+        ...
+    }
+}
+```
+Here is the code to test the `payable` function `contribute` with a `string _message` in parameter
+
+```javascript
+import "truffle/Assert.sol";
+import "truffle/DeployedAddresses.sol";
+import "../contracts/Campaign.sol";
+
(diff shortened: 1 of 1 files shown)


eggplantzzz:
Do we really want this? It seems like it might be more appropriate in the Solidity docs as it is mostly Solidity syntax. Is that right?

public/docs/getting_started/solidity-tests.md
```diff
@@ -145,3 +145,41 @@ contract TestContract {
 ```
 
 Note that Truffle sends Ether to your test contract in a way that does **not** execute a fallback function, so you can still use the fallback function within your Solidity tests for advanced test cases.
+
+Below is another example with ether transactions and parameters. Imagine a crowdfunding campaign contract on which contributors could attach a message to their contribution. Here is an extract of the contract to be tested:
+
+```javascript
```
fainashalts, line 151:
```suggestion
```solidity
```

public/docs/getting_started/solidity-tests.md
```diff
@@ -145,3 +145,41 @@ contract TestContract {
...
+    function contribute(string _message) public payable {
+        amountRaised += msg.value;
+        // todo store contribution _message in the contributions array
+        ...
+    }
+}
+```
+Here is the code to test the `payable` function `contribute` with a `string _message` in parameter
+
+```javascript
```
fainashalts, line 166:
```suggestion
```solidity
```

public/docs/getting_started/solidity-tests.md
```diff
@@ -145,3 +145,41 @@ contract TestContract {
...
+contract TestCampaign {
+    // Truffle will send the TestContract 5 Ether after deploying the contract.
+    uint public initialBalance = 5 ether;
+    Campaign campaign = Campaign(DeployedAddresses.Campaign());
+    
+    // Testing the contribute() function
+    function testContribute() public {
+        // Expected contribute increase amountRaised
+        uint256 expected = 5 ether;
+        campaign.contribute.value(5 ether) ("my test message");
```
haltman-at, line 180: Note that the `.value` syntax doesn't exist anymore in current versions of Solidity.  We probably want this to be up to date, so the syntax would be `campagain.contribute{value: 5 ether}("my test message")`.

eggplantzzz:
Since this is out of date, should we close this PR?

PR #58, opened 2017-12-29
Updating frontend/webpack tutorial to be compatible with Truffle 4

bmmpxf:
This one now has been tested against Truffle 4, with some messaging added to reflect what's changed (example: `truffle.js` [has been updated](https://github.com/truffle-box/webpack-box/commit/7c11b9355fe08527c081346e6f1e9068ac97fec7) to point to Ganache).

I also changed the commands at the end as a workaround to https://github.com/trufflesuite/truffle/issues/448, which I found via https://github.com/trufflesuite/truffle/issues/463, which happened precisely because of this tutorial. (We should still fix the underlying issue, but that's separate from this.)

### DIFF
--- a/public/tutorials/building-testing-frontend-app-truffle-3.md
+++ b/public/tutorials/building-testing-frontend-app-truffle-3.md
@@ -46,6 +46,12 @@ networks: {
 }
 ```
 
+This is the default setting for [Ganache](/docs/ganache/using), though you can change this to use any connection you'd like.
+
+<p class="alert alert-info">
+**Note**: Read more about [setting up Ganache](/docs/ganache/using). You can also use [Truffle Develop](/docs/getting_started/console), which launches a personal blockchain for testing and is built directly into Truffle, no separate installation required.
+</p>
+
 Let's get the contracts on the network:
 
 First run `truffle compile`. This will compile the `.sol` contracts into `.json` artifacts (specified in the [`truffle-contract`](https://github.com/trufflesuite/truffle-contract) library). They will appear in `build/contracts/*.json`. Now we can include contracts in our app with a simple `import` or `require` statement:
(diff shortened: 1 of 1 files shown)


public/tutorials/building-testing-frontend-app-truffle-3.md
```diff
@@ -1,55 +1,67 @@
...
 
 First run `truffle compile`. This will compile the `.sol` contracts into `.json` artifacts (specified in the [`truffle-contract`](https://github.com/trufflesuite/truffle-contract) library). They will appear in `build/contracts/*.json`. Now we can include contracts in our app with a simple `import` or `require` statement:
+
 ```javascript
 // Import our contract artifacts and turn them into usable abstractions.
 import metacoin_artifacts from '../../build/contracts/MetaCoin.json'
 ```
 
-Next run `truffle migrate`. This will deploy the contracts onto the default network running at `127.0.0.1:8545`.
+Next run `truffle migrate`. This will deploy the contracts onto the default network running at `127.0.0.1:7545`. This is the default setting for [Ganache](/docs/ganache/using), though you can change this to use any connection you'd like.
```
tcoulter: This seems to be the only mention of Ganache in this page. Should we have a section about getting Ganache up and running? If that's too much extra effort I'd say go with `truffle develop`.

bmmpxf: Depends. This was really just about making the tutorial still relevant and on-brand. I do link to the Ganache docs, though I can make that clearer.

Note that the Truffle box uses port 7545 by default. If we want to switch to use Truffle Develop, we'll need to change the box default to use port 9545. I vote for Ganache.

bmmpxf:
@tcoulter Thanks for the review!

PR #62, opened 2018-01-10
Updated outdated information

pkonyves:
* change `truffle init webpack` to `truffle unbox webpack` and it's git repo url as `init webpack` is obsolete
* the default network port is changed from 8545 to 7545. this is used by galanche and the webpack box
* added not on which file the `import metacoin_artifacts ...` code is placed
* change `truffle serve` to `npm run dev`, truffle serve runs into problems because of: https://github.com/trufflesuite/truffle/issues/448


bmmpxf:
Hi @pkonyves . We've just pushed an update to this tutorial that should address your issues here. Please take a look, and if you see something we still haven't caught, please feel free to reopen this PR. Thanks!

PR #65, opened 2018-01-13
Update solidity-testing example to use 'unbox metacoin'

cgewecke:
Resolves [truffle 702](https://github.com/trufflesuite/truffle/issues/702).


bmmpxf:
Thanks @cgewecke ! Hopefully now we've caught all the MetaCoin issues in our docs stemming from that switch.

PR #67, opened 2018-01-16
Small documentation fixes

wbt:
The first of these is a correction and the second is a clarification, so that someone doesn't make the interpretation "if I pass to link() a contract B that doesn't rely on the library being linked, the deployer will ignore B in that and any subsequent line such as the one which asks the deployer to deploy B."


bmmpxf:
Thanks for your suggestions @wbt ! More clarity is always a good thing.

PR #68, opened 2018-01-18
Add note about defining providers with functions

gnidan:
(Quick update after failing to skim the docs properly)

- Add to the code-block object type listing
- Rename header slightly for related section.


gnidan:
Danke!

PR #69, opened 2018-01-23
Update import & variable calls

propercoil:
Updated since `StandardToken.sol` was moved inside `ERC20`.


bmmpxf:
Hi @propercoil . Thanks, nice catch! Good sleuthing on the `totalSupply_` variable as well.

It's hard to keep these tutorials up to date, so we appreciate your help!

PR #73, opened 2018-01-25
Fix path to StandardToken

vincentleeuwen:


bmmpxf:
Hi @vincentleeuwen . We took care of this as part of this PR:  https://github.com/trufflesuite/trufflesuite.com/pull/69

Also, the variable was renamed, so if we use the new name, `totalSupply_`, we don't need to declare it in our contract.

https://github.com/OpenZeppelin/zeppelin-solidity/blob/master/contracts/token/ERC20/BasicToken.sol#L17

But thanks for the suggestion! If you see anything else that needs updating, please let us know.

PR #74, opened 2018-01-30
Update javascript-tests.md

nickjm:
`contract()` is not exactly like `describe()` because you cannot use `.skip` or `.only` on it, but you can use them on `describe()`.


bmmpxf:
Hi @nickjm . We actually recently added the ability to do `contract.skip` and `contract.only`. See:

https://github.com/trufflesuite/truffle-core/pull/96

It should be out when we release 4.0.7, due shortly. So I'm going to leave this content as-is, but thank you for your suggestion! Please let us know if you find anything else that needs updating.

PR #75, opened 2018-01-31
typo

nicolasburtey:

PR #77, opened 2018-02-05
Fix typo in the migrations example

joojis:


joojis:
Oops, it's a duplicate of #75, never mind.

PR #63, opened 2018-01-11
Updated example using async/await

chuacw:
Updated example using async/await, which is much easier to read.

PR #80, opened 2018-02-14
Update javascript contract testing documentations to use async await …

jaycenhorton:
…and es6

This version uses a more condensed version that does not require a pyramid of doom using chained then()s. It should be more readable and accomplishes the same thing in fewer lines.


bmmpxf:
Hey there @jaycenhorton . Thanks for this. We actually added an async/await example as part of a [different PR](https://github.com/trufflesuite/trufflesuite.com/pull/63), and it appeared to do much of the same thing. But if you see any other improvements, please don't hesitate to let us know. Thanks!

PR #81, opened 2018-02-19
The MetaCoin contract is no longer generated by default

sogasg:
Update to reflect that the MetaCoin contract is now available through Truffle Boxes.


bmmpxf:
Thanks @sogasg! I thought we had updated that MetaCoin reference everywhere, but clearly there still are some stragglers. If you find anything else out of date, please let us know!

PR #88, opened 2018-02-28
Update compile.md

diegoenriquesantos:
The text following:
To compile a Truffle project, change to the root of the directory where the project is located and then type the following into a terminal:

should be:

To compile a Truffle project, change to the project's contracts directory and then type the following into a terminal:

### DIFF
--- a/public/docs/getting_started/compile.md
+++ b/public/docs/getting_started/compile.md
@@ -8,7 +8,7 @@ With a bare Truffle [project](./project) (created through `truffle init`), you'r
 
 ## Command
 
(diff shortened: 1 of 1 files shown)


bmmpxf:
Hi @diegoenriquesantos . Actually, both are true. You can run the same commands in either the root or the contracts directory, and Truffle is smart enough to know how to find out where the files are.

We say the root directory because it's a uniform place to run commands, and sounds a little better than the technically more correct "from some directory inside your project". Though I guess we could update the text with that too.

PR #92, opened 2018-03-07
Remove extra  the in pet shop example

srt32:
Thanks!


bmmpxf:
Good catch @srt32 . Thanks!

PR #99, opened 2018-04-02
use contract name rather than filename for first metacoin example

pemulis:
This follows the directions in the migrations documentation to use contract names instead of filenames, since files may contain more than one contract. Referencing the filename still works in this example, but if I understand correctly, the tests could break if the user tries to create a contract abstraction by referencing a filename when the file has more than one contract.


bmmpxf:
Good catch @pemulis ! We keep thinking we've updated all our docs to reflect this, and folks keep finding places we haven't updated. So thank you!

PR #104, opened 2018-04-13
Update pet-shop.md

mzd555:


bmmpxf:
Hi @mzd555 . This is actually isn't correct. The `artifacts.require` function takes a contract name, not a file name, so you shouldn't need path information in the argument.

See http://truffleframework.com/docs/getting_started/migrations#artifacts-require- for more details.

But thank you for keeping an eye on things! Please let us know if you find anything else out of place.

PR #108, opened 2018-04-17
please update configuration.md

tomachinz:
I've added a section on the truffle-keys.js file as this took me a few minutes to figure out it needed to go into the ./test/ folder.


public/docs/advanced/configuration.md
```diff
@@ -264,6 +264,11 @@ dependencies: {
 }
 ```
 
+### test keys
+
+By default, Truffle looks for keys during testing in ./test/truffle-keys.js. The format of this file is seen here:
+https://github.com/ajb413/erc20-ethereum-token/blob/master/test/truffle-keys.js
```
gnidan: FYI there's nothing special about `truffle-keys.js`: it's just that when you are in a test file like that `integration.test.js`. When you do `require("./<path")`, it looks in the same directory as the test file.

I am not sure how many users are looking for this exact use case; if the docs were to include something like this, it should probably be written in as general a form as possible. By default, Truffle tests give access to the unlocked accounts, without the requirement for the keys like this. It's potentially even misleading to include this extra information.

gnidan:
Hey @tomachinz, I don't suppose you're still interested in this solution? We're looking to do issue/PR maintenance in this repo and we'd love to merge this if it's something you still want. Let us know! Thanks!

DiscRiskandBisque:
Closing this for now, but if you'd like to follow-up @tomachinz, please open a new issue!

PR #117, opened 2018-05-02
Adding Truffle quickstart

bmmpxf:
A quickstart guide has been missing from our docs for a while. This is a single page that highlights the broad concepts of a Truffle project's lifecycle, with links to the docs for more details.


bmmpxf:
Quickstart is finished! Just needs review now.

public/docs/getting_started/quickstart.md
```diff
@@ -0,0 +1,337 @@
+# Quickstart
+
+This page will take you through the basics of creating a Truffle project and migrating (deploying) a smart contract to a blockchain.
```
tcoulter: I think you should just use `deploying` here, since it's more colloquial, and then later describe that migrating/migrations are the mechanism for deployment.

public/docs/getting_started/quickstart.md
```diff
@@ -0,0 +1,337 @@
...
+
+## Creating a project
+
+To use most Truffle commands, you need to run them against an existing Truffle project. So the first step is to create a Truffle project.
+
+You can create a bare project template, but for those just getting started, you can use [Truffle Boxes](/boxes), which are example applications and project templates. We'll use the [MetaCoin box](/boxes/metacoin), which creates an [ERC20-style token](https://en.wikipedia.org/wiki/ERC20):
```
tcoulter: Did we actually upgrade the Metacoin box to be ERC20? I don't think we did.

public/docs/getting_started/quickstart.md
```diff
@@ -0,0 +1,337 @@
...
+* `test/`: Directory for test files for [testing your application and contracts](./testing)
+* `truffle.js`: Truffle [configuration file](/docs/advanced/configuration)
+
+## Exploring the project
+
+<p class="alert alert-info">
+<strong>Note</strong>: This page is just a quickstart, so we're not going to go into much detail here. Please see the rest of the Truffle documentation to learn more.
+</p>
+
+1. Open the `contracts/MetaCoin.sol` file in a text editor. This is a smart contract (written in Solidity) that generates a MetaCoin token. Note that this also references another Solidity file `contracts/ConvertLib.sol` in the same directory.
```
tcoulter: `generates` feels out of place here. Maybe `manages`? `creates` also feels nice, though it's closer to `generates`.

public/docs/getting_started/quickstart.md
bmmpxf: Oh, actually I don't know. I just assumed that MetaCoin was ERC20, as I didn't realize that one could have a token that wasn't ERC20. What should I say instead?

public/docs/getting_started/quickstart.md
bmmpxf: I'd go with `creates`, unless there's some other mechanism at work (as in, I thought the contract was creating/generating a token, but maybe not?).

bmmpxf:
Thanks for all the feedback @tcoulter . I just pushed some commits that should address everything, including a rename back to `project.md` so that no links break.

public/docs/getting_started/project.md
```diff
@@ -1,41 +1,337 @@
-# Creating a project
+# Quickstart
 
-To use most Truffle commands, you need to run them against an existing Truffle project. This section will show you how to create a new project.
+This page will take you through the basics of creating a Truffle project and deploying a smart contract to a blockchain.
 
-For those just getting started, we have many boilerplates and example applications available to you through [Truffle Boxes](/boxes). 
+## Creating a project
```
benjamincburns: What do we assume the user already knows at this point? Perhaps before we get started we should state some of these assumptions, and offer links (internal if we have them, external if not) to other prerequisite documentation?

To get started, I think we assume that a user knows

1. JavaScript
2. A bit about writing tests with Mocha/Chai
3. The gist of what a blockchain is
4. That contracts are programs ran by the Ethereum network
5. That Ethereum contracts are (most often) written in solidity
6. That there are different Ethereum networks (e.g. mainnet vs test chains)

public/docs/getting_started/project.md
```diff
@@ -1,41 +1,337 @@
...
 
-You can start with our MetaCoin example by running the following on the command line:
+To use most Truffle commands, you need to run them against an existing Truffle project. So the first step is to create a Truffle project.
 
-```shell
-mkdir MetaCoin
-cd MetaCoin
-truffle unbox metacoin
-```
+You can create a bare project template, but for those just getting started, you can use [Truffle Boxes](/boxes), which are example applications and project templates. We'll use the [MetaCoin box](/boxes/metacoin), which creates a token that can be transferred between accounts:
```
benjamincburns: We include the link to details on the MetaCoin box here, but perhaps a brief paragraph to explain what the MetaCoin box does is worthwhile?

Also, I think we unintentionally position boxes as only for those just getting started here, whereas they're beneficial for expert users as well.

public/docs/getting_started/project.md
bmmpxf: @benjamincburns Good point, a prerequisite knowledge list is a good idea. But how about we go even simpler? We have an [Ethereum Overview](http://truffleframework.com/tutorials/ethereum-overview) tutorial. Why don't we make the Ethereum tutorial the prerequisite, link to it, and make sure that all the important topics are presented there? (I believe they are, but check me.)

bmmpxf: Added a link to the Ethereum Overview tutorial. I think it's an improvement.

PR #118, opened 2018-05-02
Switched from rollup to parcel

antony:
Wasn't sure if this is the correct way to do this (or the correct place), but I've switched my box to use parcel instead of rollup, so thought this should be reflected on the website.


gnidan:
Hey @antony do you still feel this way? We'd love to consider this and are finally getting to a process of sorting through issues and PRs in this repo. Let us know!

antony:
Hey - I actually switched it back to rollup :)

This can be closed!

PR #120, opened 2018-05-07
Deprecated Solidity code

Meshugah:
Syntax for defining constructor has changed. Edit reflects this change


bmmpxf:
Hi @Meshugah . Good catch! We updated this code in a different commit, but I just went in and changed the Migrations.sol file in the related Truffle Box to match the same syntax. If you see anything else that needs updating, please let us know. Thanks!

PR #122, opened 2018-05-10
Update project.md

Panlichen:
The outputs of "Solidity test" and "JavaScript test" were mixed up.


bmmpxf:
You're right @Panlichen ! Thanks for catching that.

PR #125, opened 2018-05-15
Fix small typos and add missing words

Madisonkanna:


Madisonkanna:
Quite welcome! Thanks for all you do! @bmmpxf

PR #127, opened 2018-05-16
Update robust-smart-contracts-with-openzeppelin.md

raghunandepu:
Using contract name as function name is deprecated. I have updated line 95 (Replaced 'function TutorialToken()' with 'constructor()').


bmmpxf:
Hi @raghunandepu . I think we caught this one already, but let us know if there are still changes that need to be made. Thanks!

PR #129, opened 2018-05-17
Changed line 158 and 166

raghunandepu:
Changed line 158 : Using contract name in function is deprecated. Replaced contractor in place of function SimpleStorage.
Changed line 166 : Replaced constant with view


bmmpxf:
Thanks @raghunandepu ! We're still updating that deprecated syntax in our docs. If you find it in other places, please let us know!

PR #131, opened 2018-05-24
Updated project.md to include path to tests was not clear as previous…

lucaydaniel:
… command are from the root folder.

Either we tell the user to cd inside the test directory to run those commands or we should include the test directory in the path.

There are so many people getting confused by this ;)


bmmpxf:
Hi @lucaydaniel . We've updated this page a bit to reduce this confusion. Please take a look, and if you think it needs more work, please open this PR with additions. Thanks!

PR #132, opened 2018-05-24
FIX path to TestMetacoin.sol

pubkey:
The path in the command is wrong. This adjust the correct path.


cgewecke:
@pubkey Thanks! Nice catch.

PR #135, opened 2018-05-27
Updated to reflect correct meaning.

rafialhamd:
The phrase 'two blockchains' would mean two _blockchain networks_.
Changed to two _blockchain nodes_.


bmmpxf:
Hi @rafialhamd . That's a reasonable clarification there. Please let us know if you find any other improvements to be made. Thanks!

PR #136, opened 2018-05-31
Check windows version to determine exe or appx download of ganache

seesemichaelj:
The download link for Ganache will now default to an `exe` file; if we can detect the Windows OS version in either the `oscpu` (see https://developer.mozilla.org/en-US/docs/Web/API/Navigator/oscpu) or `userAgent`, and it's at least Windows 8 (aka >= Windows NT 6.2), then we will use the `appx` link. I'd rather the default be `exe` than `appx` to **ensure** that the user downloads something that will work.

Fixes https://github.com/trufflesuite/ganache/issues/550


seesemichaelj:
This was approved by @tcoulter and @DiscRiskandBisque in slack

PR #137, opened 2018-05-31
Add hint on URL to releases page for EXE Download

seesemichaelj:
In https://github.com/trufflesuite/ganache/issues/550 a user thought that we should say somewhere how to access an EXE download. This was mainly solved in #136, but this PR will let users know where to find the EXE if they want it over an AppX even though their OS supports it.


DiscRiskandBisque:
I'm always pro-clarity!

PR #140, opened 2018-06-03
small correction

AndreiD:
in the current version of Chocolatey adding -y will fail. It works without it, you just have to accept the license manually...


cgewecke:
@AndreiD Thanks! Could you list your chocolatey version? It might be nice to have a version reference in these instructions actually. . .

AndreiD:
now I see, you have the flag after the package name, maybe it should be in front of the package name....

PR #141, opened 2018-06-06
Correct quotes in import statement and string literal 

Purple90:
According to Solidity markup, literals and import statements are supposed to be in double quote. If a user is using a Solidity linter, it will give an error:

quotes: "openzeppelin-solidity/contracts/token/ERC20/StandardToken.sol": Import statements must use double quotes only.

quotes: 'TT': String literal must be quoted with double quotes.


bmmpxf:
Good catch @Purple90 ! Anything we can do to reduce warnings is a good thing. Though I believe this is a style suggestion and not mandatory syntax, isn't it? Either way, thank you!

Purple90:
@bmmpxf  Correct, it's not mandatory afaik. I figured for Solidity noobs like myself who use linters a lot, this might simplify things :)

Thanks!

PR #145, opened 2018-06-13
Getting started illustration should say it replays new calls rather than old ones

pcowgill:


gnidan:
[We] better get that merge conflict fixed! We'll get on that and take this to merge-town.

cds-amal:
Sorry it took so long to respond. This calls in the store have to be replayed in order to sync with potential changes.

PR #146, opened 2018-06-15
Update networks.md

lawrnce:
"When your application is run" is awkward grammar. Better choices are "When you run your application" or "As your application runs".


bmmpxf:
Hi @lawrnce. Looks good to me! Thanks for the suggestion.

PR #147, opened 2018-06-20
Remove reference to deprecated package

morenoh149:
It'd probably be better to refer to the correct command or module, but this is better than referring to a deprecated repo.


gnidan:
It sure would be nice if we didn't rename all the files.

We'll fix this merge conflict and get this into the codebase. Thanks for the time to make this change! Sorry for the delay.

eggplantzzz:
Looks like this file got deleted sometime. Sorry about taking so long with this! In the equivalent file now the package is updated to a current one. Closing this.

PR #149, opened 2018-06-24
Add v and : commands

svechinsky:
Pretty self explanatory.
They appear in the debugger but not in the online documentation


svechinsky:
Fantastic, closing.

PR #158, opened 2018-07-30
Fix warning. Added: "pragma solidity ^0.4.24;"

rafialhamd:
Add compiler version `pragma solidity ^0.4.24;` to FIX below **Warning** that appears while running `truffle test`:

> Warning: Source file does not specify required compiler version!Consider adding "pragma solidity ^0.4.24;"


eggplantzzz:
Hey @rafialhamd, sorry for letting this sit around for so long. It looks good but would you be able to resolve the conflict and update the pragma to something like `^0.5.10`? If you can do that I'll get it reviewed and merged! Thanks!

gnidan:
Looks like next step here is to figure out if this issue still exists (I wonder if those pragma statements have made into our examples by now)

eggplantzzz:
Sorry that this is now stale. I think the pragma for the equivalent file (these files got moved around) is ok for now. Closing.

PR #167, opened 2018-08-16
Correcting typo in using-an-existing-redux-store.md

skylarweaver:
Currently a typo in the instructions, where the a specific function is supposed to be referenced but only an empty set of quotes appears. Pretty confusing. This fixes that so the function you're referencing now appears.


gnidan:
Good catch! We'll fix the merge conflict and the failing build and get this merged. Thank you!

cds-amal:
This correction was already addressed.

PR #188, opened 2018-09-21
Update commands docs

eggplantzzz:
Update the docs for truffle commands.


eggplantzzz:
Should we remove the deprecated commands or leave them in?  For example, `truffle build`, `truffle serve` and `truffle watch`.

gnidan:
> Should we remove the deprecated commands or leave them in? For example, truffle build, truffle serve and truffle watch.

Actually we might need to add these commands back in. I went delete-happy but then found out that some people use these features (although I hear they're buggy)

src/docs/truffle/reference/truffle-commands.md
```diff
@@ -12,7 +12,7 @@ This section will describe every command available in the Truffle application.
 All commands are in the following form:
 
 ```shell
-truffle [command] [options]
+truffle <command> [options]
```
gnidan: Consistent semantics party 🎈

src/docs/truffle/reference/truffle-commands.md
```diff
@@ -77,15 +78,13 @@ Optional parameters:
 Helper to create new contracts, migrations and tests.
 
 ```shell
-truffle create (contract|migration|test) <ArtifactName>
+truffle create <artifact_type> <ArtifactName>
```
gnidan: Ooh this exposes a bad naming choice. We should stop using `artifact` here, so we can keep that word for specifically the JSON files in `build/contracts/`

gnidan: Without taking too much time to think of ideas, maybe we can just call it `object`? Or `entity`?

eggplantzzz: Just an idea, what if we let all these things fall under the name of artifact and we differentiate them by prefacing them with the type.  Like, 'migration artifact' or `compiled contract artifact`?  Is that too much?

eggplantzzz: Or what about 'abstraction'?

gnidan: prefixing `artifact` works. it's tough because those two words (artifact/abstraction) are both used to refer to very specific things, and they're already confusing terms. So anything that decreases risk of confusion is good 👍

PR #198, opened 2018-10-02
Update test command instruction

mehere:
truffle test command should only be run after a network is started

### DIFF
--- a/src/docs/truffle/quickstart.md
+++ b/src/docs/truffle/quickstart.md
@@ -77,7 +77,13 @@ Once this operation is completed, you'll now have a project structure with the f
 
 ## Testing
 
-1. On a terminal, run the Solidity test:
+1. On a terminal, start a test Ethereum network via [Ganache](/ganache)
+
+   ```shell
+   ganache-cli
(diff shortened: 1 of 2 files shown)


gnidan:
Thanks for this @mehere! Since this has gone about a year and a half without any attention, I'd say the next step is for someone on our side to verify that this information didn't make into our docs elsewhere. If this is still an omission, I would love to get this merged!

haltman-at:
Sorry about taking so long to get to this!  We think we're going to use the second part of this, but not the first; we don't want to recommend starting up a separate Ganache instance as the default way to use Truffle Test.

eggplantzzz:
And just to be clear on this one, you don't need to start a network separately for `truffle test` in the cases where you don't have networks named "test" or "development" in your `truffle-config.js`.

eggplantzzz:
I took this work and tweaked it and put in a more up-to-date PR [here](https://github.com/trufflesuite/trufflesuite.com/pull/198). Closing this in favor of that one. Thanks @mehere!

PR #202, opened 2018-10-04
Create an even quicker start for the built-in VM

FlashSheridan:
Create an even quicker start for debugging a test on the built-in VM, which avoids much of the complexity and discussion of alternate VMs in the quick start, and mentions a crucial step for tests on the built-in VM.  Setting up debugging was even easier than the existing Quick Start made it seem, and the discussion of alternate VMs led me through a long detour.


haltman-at:
Sorry about leaving this around for so long!  We're going to close this as we don't want to have two parallel quickstart guides, but we like the idea here, and we're opening issue #990 to continue this.  Feel free to resubmit this as a new PR to address #990!  Sorry again about leaving this for so long!

PR #204, opened 2018-10-07
Fixed params in contract adding method

igorline:
Currently it is unclear if I should be passing object containing specified properties, but actually user should be sending two values as arguments (contract config and events array)

Related to https://github.com/trufflesuite/drizzle/pull/114


chidg:
Can we please have a merge for this one? It's a one liner in the documentation, and the issue had me scratching my head until I delved into the source.

CruzMolina:
Thanks for doing this @igorline and thank you @chidg for reminding us about this! 😎

PR #206, opened 2018-10-10
add meridio img file and attach to companies on homepage

iamchrissmith:
Meridio loves using Truffle and wanted to get added to the list of companies using it.  We've include our logo file and added it to the list on the homepage (in alphabetical order).  Here is what it would look like:
screen shot 2018-10-09 at 17 43 57


tcoulter:
Hey Chris! I never saw this PR. Too bad we never acted on it. In any case, for all paying attention: we recently changed our website and locked in the logos we'll be using for now. Happy to reconsider a redesign once we get more requests!

PR #209, opened 2018-10-16
added Set Protocol logo to Awesome Companies Using Truffle

richardliang:
We at Set Protocol have been using Truffle V5 extensively and wanted our logo to be added to the list of companies on the homepage! I included a screenshot below of what it would look like:

screen shot 2018-10-16 at 3 54 31 pm


tcoulter:
Hey there. We've changed our logos significantly and have locked in the ones we want for now (locked in mostly for aesthetic reasons). Once we get more requests for logos we'll consider adding more. Thanks, and I really appreciate you using Truffle!

PR #210, opened 2018-10-21
Cleaned up test contract code

bradfordlynch:
The test contract code made use of an expected pet id and expected address in several places across test functions but repeated the definitions in multiple places thus making the test susceptible to typos. Changed the test contract code to define these once up front.


CruzMolina:
Awesome @bradfordlynch ! Thanks for doing this.

PR #211, opened 2018-10-21
Added info to explain the view keyword

bradfordlynch:
Under the section "Your second function: Retrieving the adopters" the view keyword is used in the function declaration but no explanation or link to more info was provided. I added a blurb about it and a link to the docs.


CruzMolina:
Thanks again! 🎉

PR #217, opened 2018-10-31
Use "decentralized app" instead of "distributed app"

jurglic:
Distributed and decentralized are somewhat confusingly used as
interchangeable terms, as discussed here:
https://ethereum.stackexchange.com/questions/7812.

However, most Ethereum sources including Ethereum whitepaper
(https://github.com/ethereum/wiki/wiki/White-Paper) refer to dapps as
decentralized applications.

Therefore it seems best to avoid confusion and use "decentralized apps"
here as well.


eggplantzzz:
Yes, right you are. Thanks for the addition!

PR #226, opened 2018-11-29
Update quickstart.md

ChrisChinchilla:
I'm not sure if this is the correct step to add, but the current second step assumes a file path that hasn't been explained yet, and from other tutorials I assumed to add drizzle files into a `client` folder.

This may not be the correct path, but I still feel the step is missing :)

### DIFF
--- a/src/docs/drizzle/quickstart.md
+++ b/src/docs/drizzle/quickstart.md
@@ -19,12 +19,14 @@ npm install --save drizzle
 <strong>Note</strong>: Since Drizzle uses web3 1.0 and web sockets, be sure your development environment can support these. As a development blockchain, you'll need `ganache-cli` v6.1.0+, `geth` or `parity`.
 </p>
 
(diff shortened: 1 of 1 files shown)


ChrisChinchilla:
Now I look at it, in many ways this whole quickstart is missing a few steps, say when compared to something like this post - https://truffleframework.com/tutorials/getting-started-with-drizzle-and-react. Maybe more of an issue?

eggplantzzz:
Sorry for the delay in getting to this @ChrisChinchilla, I'll try and get this reviewed!

adrianmcli:
@ChrisChinchilla I agree that this section is really barebones, but it was designed to be. I do not think the intention is to make it into a tutorial. It's more of a "how to import this into any existing modern Javascript framework".

I really appreciate you taking the effort for the PR though, but unfortunately I think I'm going to have to close this for now.

PR #241, opened 2018-12-13
Compiler config

eggplantzzz:
Update the compiler configuration docs.

### DIFF
--- a/src/docs/truffle/reference/configuration.md
+++ b/src/docs/truffle/reference/configuration.md
@@ -202,9 +202,8 @@ For more information, please see the Solidity documentation on [Compiler Input a
 
 ### external compilers
 
-For more advanced use cases, let's say your smart contract development workflow involves more than just compiling Solidity contracts. Maybe you're writing eWASM precompiles or making a two-dimensional delegatecall proxy. Or maybe you would just rather use @pubkey's solidity-cli instead of Truffle's solc configuration.
-
-This is now supported by adding a `compilers.external` object to your Truffle config:
+For more advanced use cases with artifact creation you can use the external compilers configuration.
+You can use this feature by adding a `compilers.external` object to your Truffle config:
 
 ```javascript
 module.exports = {


src/docs/truffle/reference/configuration.md
```diff
@@ -166,27 +166,123 @@ mocha: {
...
 ```javascript
-solc: {
-  optimizer: {
-    enabled: true,
-    runs: 200
+module.exports = {
+  compilers: {
+    solc: {
+      version: <string>,     // A version or constraint - Ex. "^0.5.0"
+      optimizer: {
```
gnidan: This should be inside `settings`

src/docs/truffle/reference/configuration.md
```diff
@@ -166,27 +166,123 @@ mocha: {
...
+    }
   }
 }
 ```
 
 For more information, please see the Solidity documentation on [Compiler Input and Output JSON Description](http://solidity.readthedocs.io/en/develop/using-the-compiler.html#compiler-input-and-output-json-description).
 
+### external compilers
+
+For more advanced use cases, let's say your smart contract development workflow involves more than just compiling Solidity contracts. Maybe you're writing eWASM precompiles or making a two-dimensional delegatecall proxy. Or maybe you would just rather use @pubkey's solidity-cli instead of Truffle's solc configuration.
```
gnidan: This paragraph needs links

src/docs/truffle/reference/configuration.md
eggplantzzz: Ah man I can't believe I made that mistake after all the hub bub

src/docs/truffle/reference/configuration.md
eggplantzzz: I lifted this out of the release notes because it seemed pretty thorough, I'll go ahead and keep the same links as in the notes.

eggplantzzz: @gnidan Do you think we should keep all those references in the docs or should we make it more general for the docs?

gnidan: More general would be my guess. If we want to include those references, we might just make a section "Example Projects".

WDYT?

eggplantzzz: Maybe we should just leave this more abstract.
`For more advanced use cases with artifact creation you can use the external compilers configuration.`

eggplantzzz: Then it gives some examples of different use cases below.

gnidan: Yep sounds good to me!

PR #243, opened 2018-12-13
Add new contract abstractions reference page

eggplantzzz:


src/docs/truffle/reference/contract-abstractions.md
```diff
@@ -0,0 +1,278 @@
...
+truffle(development)> const myContract = await MyContract.deployed();
+```
+
+You now have access to the following functions on `MyContract`, as well as many others:
+
+* `at()`: Create an instance of `MyContract` that represents your contract at a specific address.
+* `deployed()`: Create an instance of `MyContract` that represents the default address managed by `MyContract`.
+* `new()`: Deploy a new version of this contract to the network, getting an instance of `MyContract` that represents the newly deployed instance.
+
+Each instance is tied to a specific address on the Ethereum network, and each instance has a 1-to-1 mapping from Javascript functions to contract functions. For instance, if your Solidity contract had a function defined `someFunction(uint value) {}` (solidity), then you could execute that function on the network like so:
```
gnidan: I'm a bit wary of potentially confusing language around the term `instance`, since it might mean "abstraction instance" vs. "contract instance". 

"Create an instance" in particular seems like it might be misleading.

But maybe not a big deal. `MyContract.at()` in name is clear on its own.

PR #247, opened 2018-12-14
Add docs for autoGas and gasMultiplier

eggplantzzz:


src/docs/truffle/reference/contract-abstractions.md
```diff
@@ -153,6 +153,12 @@ const MyOtherContract = MyContract.clone(1337);
 #### `MyContract.timeout(block_timeout)`
 This method allows you to set the block timeout for transactions.  Contract instances created from this abstraction will have the specified transaction block timeout.  This means that if a transaction does not immediately get mined, it will retry for the specified number of blocks.
 
+#### `MyContract.autoGas(<boolean>)`
```
gnidan: I think this needs to be `MyContract.autoGas = <boolean>`

eggplantzzz: Good catch!

eggplantzzz: That also means some language needs to be changed

PR #262, opened 2018-12-19
Remove shell dollar sign from Solidity code block

frangio:


CruzMolina:
Good catch @frangio ! Thanks for that 😎

PR #267, opened 2018-12-21
Correction of web3.toWei() function - addressing deprecation

vvdhout:
The web3.toWei() function is deprecated as it is now stored under utils. Changed it from web3.toWei() to web3.utils.toWei()

I feel like it is important to keep these up to date as a lot of beginners will work through these docs and might pick up deprecated syntax otherwise.

Web3 Docs: https://web3js.readthedocs.io/en/1.0/web3-utils.html

All the best!


eggplantzzz:
@vvdhout Thanks for doing this and sorry for the delay. I'll get one more review and get it merged!

fainashalts:
+1 good catch @vvdhout!

PR #277, opened 2019-01-18
Update `timeoutBlocks` description as `keep waiting`

jchittoda:
Original comment
```
// timeoutBlocks: - if a transaction is not mined, keep retrying for this number of blocks (default is 50)
```
The word `retrying` is a bit misleading. As a developer might think that the Truffle will re-initiate the transaction.


CruzMolina:
Thanks for this @jchittoda!

PR #285, opened 2019-01-26
Update for new solidity compiler

FBrinkkemper:
added memory keyword


CruzMolina:
Thanks for this @FBrinkkemper !

PR #287, opened 2019-02-08
Add instructions for deploying using private key instead of mnemonic

rhlsthrm:
I created a package (forked off the truffle package) to use a private key rather than a mnemonic to deploy contracts. This works much better for me since I can use my MetaMask account's private key to deploy. I use this all the time, so it would be useful to add it to the official truffle docs.

Please let me know if there are other places I can add this reference as well.


eggplantzzz:
Sorry for the delay in getting to this @rhlsthrm but truffle-hdwallet-provider already allows you to do this. Check out https://github.com/trufflesuite/truffle/tree/master/packages/truffle-hdwallet-provider#private-keys

rhlsthrm:
@eggplantzzz yep, they have essentially made this package obsolete!

PR #294, opened 2019-02-25
Fix grammar

chriseth:


CruzMolina:
Thanks for that @chriseth ! 🍬

PR #303, opened 2019-03-07
Removed note to pull docker manually

daniellehrner:
Removed outdated note about using docker as compiler option


CruzMolina:
Thanks @daniellehrner !

PR #311, opened 2019-03-13
Content edits on trufflecon2019.html

shayzee:
Updated the copy and CTA for speakers and removed old info for 2018 section


DiscRiskandBisque:
Note: Closing because I believe we resolved this on a Zoom and the page has since been replaced by your new design.

PR #315, opened 2019-03-14
Show docs sidebar before loading the page instead of after

seesemichaelj:
This PR will add the `show` class server-side instead of client-side. This prevents a jarring "it's open, then we close on load, then it opens again" UX.


DiscRiskandBisque:
LGTM!

PR #317, opened 2019-03-15
Remove unused code and improved description.

abax1:
Removed `generateStore` from `import { Drizzle, generateStore } from "drizzle";` as it is not used.


adrianmcli:
Thanks!

PR #328, opened 2019-03-27
Fix browsersync on windows by removing full path

tcoulter:
According to [this page](https://github.com/paulmillr/chokidar/issues/668#issuecomment-357235531), the full path is technically an invalid glob (hey, I don't make the rules). 

> If you want to use globs, you must ensure that only forward slashes are used as path separators on your input to chokidar.

Can non-Windows users test?

### DIFF
--- a/scripts/dev.js
+++ b/scripts/dev.js
@@ -10,10 +10,10 @@ app(false).use(browserSync({
     }
   },
   files: [
-    path.join(__dirname, '../helpers/**/*'),
-    path.join(__dirname, '../layouts/**/*'),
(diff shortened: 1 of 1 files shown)


adrianmcli:
Still works as expected on MacOS.

eggplantzzz:
@tcoulter @adrianmcli Is this PR mergeable?

tcoulter:
It somehow got merged through other means. Looks like  you can blame it on me actually: https://github.com/trufflesuite/trufflesuite.com/blame/6d1494148462ddf286038d82b01bd71a371f7d90/scripts/dev.js#L13

Closing.

PR #343, opened 2019-04-07
Removing code element not in example codebase causing error

folkengine:
See [here](https://github.com/trufflesuite/drizzle-event-demo/blob/master/app/src/middleware/index.js) for the actual example code. 

Thanks for all the great work!


cds-amal:
Thanks for your PR contribution @folkengine!

PR #382, opened 2019-05-10
Removing self closing tag for DrizzleProvider

antododo:
`DrizzleProvider` includes `LoadingContainer` and `MyContainer` so it can't be self closing.

ps: I don't know why github show the last line as it has changed. I didn't touch it and it still exactly the same 🤔


antododo:
@cds-amal Did you have time to review this PR? Let me know if I need to change something.

PR #391, opened 2019-05-15
Doc update: deploying using array syntax is deprecated

thecryptofruit:
Referencing issues #1136 and #1139.


CruzMolina:
Ah, thanks for pointing this out @thecryptofruit. Will mull this over. I think we might just want to remove the reference to array syntax deployment completely. Quorum migrations still allow this because they use the legacy migrations system, but we are probably going to eventually transition it into the new system.

eggplantzzz:
I think we probably do want to just completely remove that whole blurb about deploying with array syntax. If you can edit this PR @thecryptofruit, I will try and get it reviewed and merged! Thanks!

eggplantzzz:
Merging!

Issue #400, opened 2019-05-21
Duplicate Wrapper issue on Truffle Box Detail Page


joshmobley:
The detail view of truffle boxes seems to double-wrap the contents. You'll notice the double header, and the scrollable portion of the page is smaller than the viewport.

This was seen on Chrome 73

![image](https://user-images.githubusercontent.com/933565/58109941-8cee3400-7bbc-11e9-8ee9-28f814586e8e.png)

![image](https://user-images.githubusercontent.com/933565/58109950-9081bb00-7bbc-11e9-9571-dd77173d6c08.png)

joshmobley:
I'm going to take a pass at a fix as well, if you'd like to assign me.

joshmobley:
Fixed with #401

PR #401, opened 2019-05-21
removing duplicated meta/head code that way breaking layout

joshmobley:
PR for issue #400 

Duplicated head markup was causing layout issues on the truffle box detail pages.

### DIFF
--- a/layouts/box-single.hbs
+++ b/layouts/box-single.hbs
@@ -66,66 +66,7 @@
           {{/if}}
         </div>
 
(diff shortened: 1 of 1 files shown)


eggplantzzz:
Can you please give a link to where the problem exists and maybe some more detail about what you are viewing it with etc.?

joshmobley:
More details are in the issue #400

eggplantzzz:
Thanks @joshmobley! I'll go ahead and take a look into this.

PR #421, opened 2019-06-12
Fix troublesome typo in example of displaying Drizzle data from the store

leopoldjoy:


adrianmcli:
Good catch @leopoldjoy, thanks!

PR #425, opened 2019-06-14
Add warning about NodeJS 12

filips123:
I've added warning about NodeJS 12. This is related to https://github.com/trufflesuite/truffle/issues/2070#issuecomment-501699184.

### DIFF
--- a/src/docs/truffle/getting-started/installation.md
+++ b/src/docs/truffle/getting-started/installation.md
@@ -10,8 +10,10 @@ npm install -g truffle
 
 ## Requirements
 
-* NodeJS v8.9.4 or later
+* NodeJS v8.9.4 - v11.15.0
 * Windows, Linux or Mac OS X
+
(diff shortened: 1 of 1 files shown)


Karunamon:
Heya.. could this be merged? I just ran into this as well and now have to reinstall a whole new node version. Might save the time of the next clueless newbie to come along :)

eggplantzzz:
I'm not sure what we should do about this. In general, the only problem I am aware of with Node 12 is that (because of an issue with Mocha) you can't run `truffle test` in the Truffle developer console. Other than that Truffle is usable with Node 12.

DiscRiskandBisque:
@eggplantzzz We could add a note to the [Testing Your Contracts](https://www.trufflesuite.com/docs/truffle/testing/testing-your-contracts) Page to a similar effect?

eggplantzzz:
@DiscRiskandBisque It looks like it might not even be worth it now since [this](https://github.com/trufflesuite/truffle/pull/3255) is on its way through which solves the issue here.

haltman-at:
Closing this as Node 12 (and Node 14) is now supported.

PR #431, opened 2019-06-20
Update pet-shop tutorial with new imagery and ES6 syntax

Elaniobro:
#### Description: ✍️

Updated `pet shop` tutorial's metamask img for network name to reflect the new metamask UI.

##### Before:
image
##### After:
metamask-customrpc


Elaniobro:
May we leave this open for now? (I can't apply labels)

I want to go through and verify some of the other screenshots in the tutorial. 

I may also opt to update the code to use ES6 javaScript syntax as well.

eggplantzzz:
Sorry for the delayed response @Elaniobro. Sure, we can leave this open for now, keep me updated on your progress. Thanks for the help!

Elaniobro:
@eggplantzzz I went ahead and updated the javaScript from jQuery -> ES6, we can move forward with the review, I will also re-work the tutorial from start to finish to capture and image updates as well.

eggplantzzz:
Thanks @Elaniobro! However, this job may be bigger than you think. If we do this, then we will also have to update the [pet shop box](https://github.com/truffle-box/pet-shop-box) to be set up to handle ES6 stuff. I haven't really thought too much into what kind of specific setup this would require but I just wanted to note it here before you got too far along.

Elaniobro:
@eggplantzzz good point, did not consider that. Where might I find the source code for said box? looking under the hood will help me understand how big this update is/would be..

eggplantzzz:
I linked it above in my comment.
Here https://github.com/truffle-box/pet-shop-box

If I get around to spec'ing out what it might entail I'll leave some more info here!

Elaniobro:
@eggplantzzz wow, that link completely went over my head. Thanks for the long form version as well. I'll take a look at it and take a stab at it while you prepare some specs. 👍🏽

eggplantzzz:
Ugh, sorry @Elaniobro, this got left behind again :( I think this work might still be useful though. Unfortunately the target file has changed. I think it is `src/tutorial.md`. 

On revisiting this, it seems like there might not be any changes necessary for the box code. Also I think we probably should keep the jQuery stuff where it is for now so we don't have to deal with another dependency. There are also a lot of code formatting issues spacing-wise. I don't know, it's been a long time but if you still are interested I think we can get this pushed through. Thoughts?

Aadhinana:
The connection API to metamask also needs an update?
[The new Metamask Update docs](https://docs.metamask.io/guide/getting-started.html#basic-considerations)

BEFORE: 
```js
// Modern dapp browsers...
if (window.ethereum) {
  App.web3Provider = window.ethereum;
  try {
    // Request account access
    await window.ethereum.enable(); // This has been deprecated in favour of  ethereum.request
  } catch (error) {
    // User denied account access...
    console.error("User denied account access");
  }
}
```
AFTER: 
```js
// Modern dapp browsers...
if (window.ethereum) {
  App.web3Provider = window.ethereum;
  try {
    // Request account access
    await window.ethereum.request({ method: "eth_requestAccounts" }); // New API
  } catch (error) {
    // User denied account access...
    console.error("User denied account access");
  }
}
```

gnidan:
Thank you @Elaniobro for doing this, but unfortunately we've completely dropped the ball on this and now this PR is too out of date for us to proceed :(. Closing this in favor of alternate pet-shop revitalization efforts.

PR #450, opened 2019-07-06
truffle test: Describe the exit status

icherkashin:
Per https://github.com/trufflesuite/truffle/issues/2155

Truffle documentation does not describe the exit status of `truffle
test` command. Neither does `mocha`, except in its source code.

The changes introduced by this commit may benefit the project in these
ways:
  - Providing a link to `mocha` source code shows that the issue is not
  caused by `truffle`
  - Users searching for `mocha exit status` may find the answer in
  `truffle` documentation, which increases awareness of the project

### DIFF
--- a/src/docs/truffle/reference/truffle-commands.md
+++ b/src/docs/truffle/reference/truffle-commands.md
@@ -298,7 +298,13 @@ Run JavaScript and Solidity tests.
 truffle test [<test_file>] [--compile-all] [--network <name>] [--verbose-rpc] [--show-events]
 ```
 
-Runs some or all tests within the `test/` directory as specified. See the section on [Testing your contracts](/docs/getting_started/testing) for more information.
+Runs some or all tests within the `test/` directory as specified.
+
+See the section on [Testing your contracts](/docs/getting_started/testing) for more information.
+
+Exit status:
+
+* Per mocha's [source code](https://github.com/mochajs/mocha/blob/5d4dd98747637d0e7ed3007328ec9627dd7eda41/lib/cli/run-helpers.js#L35): typically, the exit status is the number of failed tests or 255 if there has been more than 255 failed tests.
 
 Options:
 


eggplantzzz:
Thanks @icherkashin, let me get this reviewed.

eggplantzzz:
@icherkashin Where is the exit code exposed by Truffle? Can you give me some steps that will demonstrate how Truffle exposes it?

yxliang01:
@eggplantzzz I don't really understand your question. So, I was the one raised up this problem (lack of documentation of exit code). By "exit code", we mean the exit status `truffle` (the CLI) informs the system its status at the time the process terminates.

eggplantzzz:
What I mean is why does this need to be documented? Does Truffle expose this somehow? In other words, if mocha provides an exit code when it is done but then Truffle doesn't do anything with it, then it doesn't seem to be useful to document it. Like, if Truffle doesn't do anything with the exit code then it is not useful to users.

If you know where this exit code is used by Truffle then let me know!

DiscRiskandBisque:
Hi @icherkashin and @yxliang01. We're going through and catching up on issues and PRs with the webiste--would love to get your thoughts on @eggplantzzz question above so we can determine next steps.

eggplantzzz:
So currently I don't see how this would be useful to document. While running `truffle test`, Truffle doesn't really do anything with the exit code from Mocha and it doesn't expose it in any way for people to use. So for that reason I don't think it makes any sense to document it. Please let me know if I'm mistaken so we can figure out the most sensible thing to do!

yxliang01:
Hi, indeed, I think the exit code of the `truffle` CLI command is from mocha. In my projects, the exit code is helpful in determining the test suite execution status. While truffle doesn't do anything with it, ideally we shouldn't expect truffle users to refer mocha documentation to understand the `truffle` CLI exit status? I think it is better to treat truffle as a standalone tool that is based on mocha, meaning that the documentation should be self-contained.

yxliang01:
By the way, following code snippet shows how we can access the exit status with Bash. I hope it answers the previous question.

```bash
truffle test
echo $?
```

eggplantzzz:
Hey @yxliang01, I don't think that is correct. [Here](https://github.com/trufflesuite/truffle/blob/develop/packages/core/cli.js#L52-L107) is the top-level view of how a command is run and then an exit code set. I think Truffle internally uses the mocha exit code as you can see [here](https://github.com/trufflesuite/truffle/blob/develop/packages/core/lib/test.js#L159-L162) but that is not the actual exit code for Truffle.

If you think I'm mistaken could you show me in the code where you think it is exposed to users? Thanks!

haltman-at:
Hi, sorry for leaving this for so long!  We're going to close this but we've opened issue #991, and we'll fix it there; feel free to submit a new PR for that.  Thanks for bringing this up, sorry again about leaving it so long!

PR #529, opened 2019-09-10
Move DLT integration docs to its own section

seesemichaelj:
We discussed about doing this on slack back in March but never did it. The final outcome came from @CruzMolina:

>my thoughts are, it's dev documentation, so whatever aligns w/ long term goals (multi-ledger support) while also being easy for truffle users new & old to navigate seems to make the most sense.
>I tend to think something like "Distributed Ledger Support" fits that bill

This PR does that:

image

I could not find any URL references that needed updating in the website or truffle repo. I added in this PR a redirect plugin which will redirect the old links to the new links


seesemichaelj:
Seems like it's failing because we're using NodeJS 6.9 in our travis builds :scream: and the dependency I am using uses the spread operator (which isnt supported in v6). we really should update to 10. i can ear mark that into this PR?

seesemichaelj:
@CruzMolina updated the travis file and now gives a successful build

PR #532, opened 2019-09-15
Fixing broken images after switch to monorepo

KorbinianK:
Some of images in the documentation were not updated after moving Drizzle to a mono repository.


eggplantzzz:
@KorbinianK Thanks for fixing those!

PR #577, opened 2019-12-05
Add special methods docs

eggplantzzz:
Add docs for the `estimateGas` and `sendTransaction` methods found on Truffle contract objects


eggplantzzz:
@haltman-at I added mentions to the 2 methods you suggested and re-organized everything to fit. Thanks!

PR #617, opened 2020-03-06
Explain avoiding recompile with debugger

haltman-at:
Adding some documentation about how to avoid a recompile on debugger startup with the latest Truffle version.

### DIFF
--- a/src/docs/truffle/getting-started/debugging-your-contracts.md
+++ b/src/docs/truffle/getting-started/debugging-your-contracts.md
@@ -94,6 +94,12 @@ $ truffle debug
 
 Regardless of how you start the debugger, once it is running you are not limited to debugging only the transaction you launched it with; it is possible to unload the current transaction and load a new one, as described below.
 
+NOTE: If your project was not compiled all at once (or under certain other
+conditions), the debugger will have to do its own compile of your project on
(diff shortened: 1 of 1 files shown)


src/docs/truffle/getting-started/debugging-your-contracts.md
```diff
@@ -94,6 +94,12 @@ $ truffle debug
 
 Regardless of how you start the debugger, once it is running you are not limited to debugging only the transaction you launched it with; it is possible to unload the current transaction and load a new one, as described below.
 
+NOTE: If your project was not compiled all at once (or under certain other
+conditions), the debugger will have to do its own compile of your project on
+startup.  This can be very slow.  If you compile your whole project at once,
+however, the debugger can likely avoid the initial recompile, speeding up
+startup greatly.
```
gnidan: Note on `NOTE:`: since this is getting rendered as hypertext, probably better to find an example use of one of those info boxes instead, so it can be in a blue or yellow box or whatever.

haltman-at: Good idea!  Just pushed a revised version.

seesemichaelj:
FYI @gnidan this looks like it's ready for a re-review (I'm cleaning up the trufflesuite.com branches)

PR #624, opened 2020-03-17
Add a query parameter to the css link to fix cloudflare caching

seesemichaelj:
We use cloudflare CDN caching to speed up page loads; this is great and all except for when static files are updated. This is primarily geared towards CSS changes. I add a `?${Date.now()}` to the index.css link at build time to combat this

### DIFF
--- a/layouts/blog-post-single.hbs
+++ b/layouts/blog-post-single.hbs
@@ -50,7 +50,7 @@
 
   <link href="https://fonts.googleapis.com/css?family=Grand+Hotel|Open+Sans|Oswald|Varela+Round|Roboto+Condensed|Roboto+Mono" rel="stylesheet">
   <link rel="stylesheet" href="https://use.fontawesome.com/releases/v5.0.9/css/all.css" integrity="sha384-5SOiIsAziJl6AWe0HWRKTXlfcSHKmYV4RBF18PPJ173Kzn7jzMyFuTtk8JA7QQG1" crossorigin="anonymous">
-  <link rel="stylesheet" href="/css/index.css">
+  <link rel="stylesheet" href="/css/index.css?{{ cssVersion }}">
 </head>
 <body>
   {{> ga-tag-body}}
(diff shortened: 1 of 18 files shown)


seesemichaelj:
@davidmurdoch I'm going to go ahead and merge this as I need this for docs we need accessible tomorrow morning. However, please review this anyway and we can apply changes then.

davidmurdoch:
This cache busting technique opens us up to an attack called "cache poisoning". A competitor could easily request our CSS files from cloudflare for ever future version, which would poison the cache with stale data. That said, time-based cache busting as you've done here is a bit better than incremental version based cache-busting, as it'd take a ton of requests to successfully poison the cache.

Another issue: the way it's implemented here invalidates the cache for files that haven't changed. Which isn't a huge deal, but there are better ways.

A better version would to hash each file's contents (md5 is fine for this) and use that as the cache bust param.

Finally,  some buggy proxies, CDNs, and old browsers don't properly invalidate the cache when the query string changes. I've personally never actually seen this be a problem in the wild, and I think we'll be safe to continue using query strings here, as the alternative (change the file name itself) may be a bit cumbersome to implement with our static site generator.

seesemichaelj:
> the way it's implemented here invalidates the cache for files that haven't changed. Which isn't a huge deal, but there are better ways.

Unfortunately with our current setup, we only have one `index.css` file that is **all** of the css for the entire website, and it is used for every html page

> A better version would to hash each file's contents (md5 is fine for this) and use that as the cache bust param.

with that said, hashing does seem more appropriate. it helps tremendously with cache poising as well as prevent `index.css` from being cache busted if it doesn't change

> I think we'll be safe to continue using query strings here, as the alternative (change the file name itself) may be a bit cumbersome to implement with our static site generator.

i agree with this sentiment

seesemichaelj:
thanks for the info/tips!

seesemichaelj:
I started looking at something like: https://gist.github.com/seesemichaelj/88b6c942e951e98191359666e0260d52

however, the step in which we need to inject data into handlesbars doesnt seem to happen at a point after sass runs. i'm not sure if there is an easier way here without spending much more time on this

davidmurdoch:
Hm, yeah, we'll may need to handle this with via our in `use` that wraps the sass one, injecting our own metadata somehow.

Ooorrrrr, we could just hash the modified time of the last modified SASS file and hope some nefarious person doesn't read this comment and discover my brilliant (horrible) cache poisoning prevention scheme. 😅

seesemichaelj:
Hm well I think I'm reaching my amount of effort on this one

PR #643, opened 2020-04-16
Don't re-define `contract` test helper in example

paulperegud:
Copy-pasting this example breaks `contract('Name..', accounts async => {..});` test helpers. Beginner will waste a lot of time trying to find where is the problem - like I did.


eggplantzzz:
I don't really think we want to make this change. It has become a convention to use `contract` as the variable name for the exported function from `@truffle/contract`. While in a test, you should not be using `@truffle/contract` directly. Truffle provides the `artifacts.require` method to obtain a reference to your contract abstractions.

Looks like we do need to update the library from `truffle-contract` to `@truffle/contract` though...

paulperegud:
I can't really answer you what is the best way of going forward with this problem. I don't have JS expertise - otherwise I wouldn't have walked into the problem. Warning users in docs is better than nothing, IMHO. The real problem is JS itself - it should want about shadowing of variables.

eggplantzzz:
Thanks @paulperegud, yeah JS can be wonky. Well if you happen to spot a good place in the docs for this warning then let us know and we'll add it or gladly discuss another PR!

PR #647, opened 2020-04-18
Updated the correct way to install drizzle

srinivasmangipudi:
I found this in one of the git issues and believe is the correct way to install drizzle. the way mentioned in the tutorial through scrypt errors for node 12 etc.


eggplantzzz:
Thanks for doing this! Let me get one more approval.

PR #649, opened 2020-04-23
Add section on source code verification

rkalis:
Today on Twitter someone complained that source code verification with Truffle + Etherscan was a pain. @gnidan replied to try truffle-plugin-verify, so I suggested adding it to trufflesuite.com for awareness.


rkalis:
Done.

PR #660, opened 2020-05-04
Update references to truffle-contract

eggplantzzz:
Update references to the 'truffle-contract' library to use '@truffle/contract' instead.

There are some more references to `truffle-contract` [here](https://github.com/trufflesuite/trufflesuite.com/blob/develop/src/blog/truffle-v5-has-arrived.md) and [here](https://github.com/trufflesuite/trufflesuite.com/blob/develop/src/tutorials/upgrading-from-truffle-2-to-3.md) which we could consider changing as well but they are not really relevant any more. I'm not sure what we should do for those.


DiscRiskandBisque:
@eggplantzzz Good point about the older references. We need to discuss what we're going to do with outdated content. Likely could put a warning at the top and be good.

PR #693, opened 2020-06-23
Replace Markdown links in HTML with HTML links

haltman-at:
This PR replaces the Markdown links inside HTML on the debugger documentation page with HTML links, so that they'll work.


src/docs/truffle/overview.md
```diff
@@ -5,7 +5,9 @@ layout: docs.hbs
 <img style="max-width: 160px;" src="/img/truffle-logo-dark.svg" alt="Truffle Logo" />
 
 <div class="text-center docs-badges">
-  [![Join the chat at https://gitter.im/consensys/truffle](https://badges.gitter.im/Join%20Chat.svg)](https://gitter.im/consensys/truffle?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge)
+  <a href="https://gitter.im/consensys/truffle?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge">
```
seesemichaelj, line 8: can you actually fix this in the other `overview.md` files while you're at it?

haltman-at:
This is actually not done yet!  I put this up before realizing the extent of the problem.  Will keep finding more and re-review.

haltman-at:
> What grep command did you use to find these (while not finding all markdown links)?

It was mostly manual, TBH...

src/docs/truffle/overview.md
haltman-at: Done!

PR #706, opened 2020-07-06
Update quickstart.md

rattle99:
`npm install --save drizzle` gives an error.


rattle99:
> Ah this is a great find! Thanks!

Welcome!

PR #717, opened 2020-07-09
Event updates

kevinbluer:
Removing yesterday's Blockchain Academy Mexico event and adding the upcoming Kaleido webinar. @CombsieZ I've used the thumbnail Kaleido gave us...thoughts welcome if this works from a style perspective.


DiscRiskandBisque:
LGTM--thanks for rebasing to `develop` @kevinbluer !

PR #743, opened 2020-07-29
Update the text for the unbox command description

eggplantzzz:
List all supported formats that you can use when specifying a box name.

### DIFF
--- a/src/docs/truffle/reference/truffle-commands.md
+++ b/src/docs/truffle/reference/truffle-commands.md
@@ -366,7 +366,7 @@ Options:
 
 Also note you can add a `#` followed branch name to the end of all of the
 above formats to unbox from a specific branch - like
-truffle-box/bare-box#myBranch
+"truffle-box/bare-box#myBranch"
 
 ### version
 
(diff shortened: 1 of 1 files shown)


src/docs/truffle/reference/truffle-commands.md
```diff
@@ -341,18 +341,32 @@ Options:
...
+defaults to the current working directory if this argument is not provided.
 
-You can also design and create your own boxes!  See the section on [Truffle boxes](docs/truffle/getting-started/truffle-boxes) for more information.
+See the [list of available Truffle boxes](/boxes).
+
+You can also design and create your own boxes!  See the section on [Truffle boxes](docs/truffle/advanced/creating-a-truffle-box) for more information.
 
 Options:
 
 * `<box_name>`: Name of the Truffle Box. (required)
```
eggplantzzz, line 356: Maybe this should be removed from the options section as it isn't really an option.

DiscRiskandBisque: That's a good point. Same applies to `run` and likely other commands on that page. Will merge this for now.

PR #775, opened 2020-08-31
added warning about conflict between contract names and node objects

mktcode:
referring to: https://github.com/trufflesuite/truffle/issues/3329

image

### DIFF
--- a/src/docs/truffle/getting-started/using-truffle-develop-and-the-console.md
+++ b/src/docs/truffle/getting-started/using-truffle-develop-and-the-console.md
@@ -10,7 +10,7 @@ Sometimes it's nice to work with your contracts interactively for testing and de
 * **Truffle Develop**: An interactive console that also spawns a development blockchain
 
 <p class="alert alert-danger">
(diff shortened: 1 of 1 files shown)


eggplantzzz:
Thanks again @mktcode! Let me take a look and get back to you.

eggplantzzz:
I have 2 comments about this. I feel like this should be a **Note:** rather than the red warning box.

I feel like the messaging should be slightly different. I think we should try to communicate 2 things to people:
  - your contract names get loaded as variables into the console context
  - for this reason avoid names that might conflict with Node native objects like Buffer or String

What do you think?

mktcode:
What about a note regarding a future release? That was the whole point for me because I was told that a fix for this issue, by implementing some namespacing mechanic, will probably not happen until the next major release. So I thought it might be a good idea to put a note in the docs.

> Your contract names get loaded as variables into the console context. For this reason it is recommended to avoid names that might conflict with Node's native objects like Buffer or String. For updates see the related [issue on GitHub](https://github.com/trufflesuite/truffle/issues/3329).

Better?

eggplantzzz:
I like it! I think that should be helpful.

mktcode:
Updated. :)

eggplantzzz:
Thanks @mktcode!

eggplantzzz (approved):

cds-amal (approved):
LGTM!

mizz75 (commented):
Hi

PR #778, opened 2020-09-02
feat: minimize image size within blog

AemieJ:
I observed that the images take a lot of space within the blog section in smaller screen width. Thus, I have proposed the following changes. It's a minor change but gives a better outlook for smaller screens.

### Before
before-blog-image-truffle

### After
after-blog-image-truffle

### DIFF
--- a/src/sass/pages/blog.scss
+++ b/src/sass/pages/blog.scss
@@ -27,6 +27,13 @@
   border: 4px solid #3fe0c5;
 }
 
+@media screen and (max-width: 767px) {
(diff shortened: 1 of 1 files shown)


eggplantzzz:
@AemieJ Thanks for the proposal! I think this looks like a definite improvement, let me get @DiscRiskandBisque to take a look. Thanks again!

DiscRiskandBisque (approved):
Thanks @AemieJ! This is a nice improvement.

PR #815, opened 2020-10-26
Switch out link to hopin directly

CombsieZ:
Please double check that the open in new window tag is working properly. I think my browsers weren't allowing it to open new windows.


DiscRiskandBisque:
The issue was the `.scroll-link` class on the signup buttons. The scrolling script uses that class to override the default link behavior and initiate the scrolling. By removing it, the default link behavior is restored.

DiscRiskandBisque (approved):

PR #817, opened 2020-10-27
Add blog for using ethpm & truffle

njgheorghita:
Tutorial style blog for how to use ethpm & truffle.


eggplantzzz:
I went ahead and added a necessary json snippet, rebased, and went through the installation portion of the tutorial here. Looks good but I had a couple of comments that I think would make this smoother.

1. You may want to point out that it is necessary to have a ropsten configuration to use this. The error does not make the solution completely obvious when you don't have it configured.
2. You may also want to note that you need ENS enabled. Relevant docs are [here](https://www.trufflesuite.com/docs/truffle/advanced/ethereum-name-service).
3. The erc20 package that you reference in the install section does not work out of the box so I would think that you should use the `safe-math` package (or some other alternative that works) which worked for me when I went through.

So I think we're on track for releasing this Wednesday! Let me know about the above.

eggplantzzz:
Oh also do note that I force pushed so you'll have to sync your branch back up!

njgheorghita:
@eggplantzzz Thanks for the review & feedback! What exactly was wrong with the `erc20` package. Was there trouble installing it or using it?

eggplantzzz:
@njgheorghita It will not install. Here is a screenshot of what I get 
Screen Shot 2020-11-03 at 12 51 08 PM

njgheorghita:
@eggplantzzz Thanks for the output. It looks like I linked the incorrect registry inside the `truffle-config.js`. Would you mind testing this one more time? `erc20-example` is a much better example of what's possible rather than `safe-math`. I've updated all of the examples in the sample `truffle-config` to be equal to their default values. You should be able to install `erc20-example` with or without the sample `truffle-config` ethpm settings.

DiscRiskandBisque:
@eggplantzzz Is this good to go?

eggplantzzz:
Not yet @DiscRiskandBisque, we are going to wait until releasing 5.2.0 which is kind of in a holding pattern right now. I'll get back to you here when it is ready to roll out. Thanks!

eggplantzzz:
Integrating this with the ethpm docs PR and continuing this [here](https://github.com/trufflesuite/trufflesuite.com/pull/889)

PR #841, opened 2020-11-11
Elaborate on artifact json naming convetions.

MrClottom:
Don't know if this extra notice is necessary, it's just that running `truffle test` wasn't working for some reason despite my source file being named correctly. My initial understanding was that artifact files are named after their source so I was sure that a typo wasn't the error. After a lot of debugging I realized that my contract was named differently and that that was the issue. In hindsight this fact is mentioned in the `running migrations` section of the docs but I feel I could've avoided the error if it was already mentioned in the `Compiling contracts` section.

*`TL;DR`* Despite this detail being mentioned in 'Running Migrations' I believe it should be mentioned again here since developers simply compiling and testing contracts may not necessarily deem it relevant to their issue to check in the 'Running Migrations' section.


DiscRiskandBisque:
I'm all for extra clarity @MrClottom and thank you for the edit!

DiscRiskandBisque (approved):

PR #864, opened 2020-12-02
Document that in-test debugging currently doesn't work for reverted transactions

haltman-at:
Redoes #863 with correct branch name this time. :)

Fixes https://github.com/trufflesuite/truffle/issues/3543.

PR #865, opened 2020-12-03
Replace use of Markdown ** inside HTML with <strong> tags

haltman-at:
This PR is something of a continuation of #693; that was me noticing various links not working due to them being Markdown links inside HTML, here it's the same thing except I've replaced `**text**` with `<strong>text</strong>` when inside other HTML, because again the former doesn't work inside other HTML.

Note I didn't do a thorough manual sweep like I did with the links; I just searched for a few particular patterns I noticed as recurring problems.

Also, I added warning or info boxes around any other `**Note**:`s that I found, including my one from #864 (oops).  However there's one particular one I left out (the one on box names) because I was having a hard time getting it to work inside an info box; no idea why.  Oh well.


gnidan (approved):
Sweet!

PR #867, opened 2020-12-05
replace getBalance with balanceOf

zacel:
In this documentation getBalance is used to get balance of a token. I suggest changing this to balanceOf as that is the function for getting balances from erc20 tokens which is likely a primary use base for tokens that may be tested like this.


eggplantzzz:
Unfortunately this won't work "as is" as the example is based off of Truffle's "metacoin" box. In that box the method name is `getBalance`. So if we really wanted to do this we would also have to change the method name in the Truffle box.

zacel:
ah that makes sense thanks for explaining it

PR #896, opened 2021-01-20
Update ethpm example package

njgheorghita:
Updated the example package
- removed build dependencies to simplify the example
- fixed import paths in source contracts


njgheorghita:
@eggplantzzz Heyo! Here's the updated package. Everything seems to be working when I've tested it locally, but let me know if there are any more snafus or questions.

eggplantzzz:
Ah @njgheorghita, sorry for the delay I missed this. Thanks for updating this!

PR #909, opened 2021-02-11
Fix markdown to display hyperlinks for ganache docs

fainashalts:
This PR changes the `p` tag to be a `div` and puts a space below the opening div tag in order for the hyperlinks to ganache to render properly.

### DIFF
--- a/src/tutorials/chain-forking-exploiting-the-dao.md
+++ b/src/tutorials/chain-forking-exploiting-the-dao.md
@@ -1,6 +1,7 @@
-<p class="alert alert-info">
+<div class="alert alert-info">
+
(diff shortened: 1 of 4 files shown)


fainashalts:
@DiscRiskandBisque I'm not sure if this is the best way to do this but it seems fine? All of these tutorials will need to be archived I believe, so this is just a temporary change until that gets sorted out!

DiscRiskandBisque:
LGTM thanks @fainashalts ! Only thing you'll see I did was remove the padding from the alert paragraphs.

DiscRiskandBisque (approved):

PR #910, opened 2021-02-11
Update Tezos support documentation for delphinet

bergarces:
This PR updates Carthagenet network mentions to Delphinet.

It also provides some clues for the user to understand that a different test network can be used if Delphinet is disabled at some point.


DiscRiskandBisque (approved):
Thanks @bergarces , looks good!

PR #911, opened 2021-02-11
Fix unterminated string in require statement

wpapper:
Contributing a small typo fix in the documentation around using `@openzeppelin/truffle-upgrades`.

Happy to make any requested changes, I have read the contribution guidelines. I assumed that the branch should be `content` rather than `fix` because the issue is with a code sample inside of a blog post, rather than any of the code used by the website itself.


DiscRiskandBisque:
Uh oh! 😅 Thanks for this @wpapper .

PR #938, opened 2021-03-25
5.2 blogpost

eggplantzzz:


src/blog/features-from-version-5.md
```diff
@@ -0,0 +1,110 @@
...
+
+### Experimental Vyper debugging
+
+You can now use Truffle Debugger with Vyper!  Support for this is still in its early stages; you won’t be able to inspect variables, I’m afraid.  But you can step through a Vyper transaction, and there will likely be more in the future!
+
+But perhaps the biggest improvement to the debugger is...
+
+### Debugging verified external contracts
+
+Use the `--fetch-external` (or `-x` for short) option with Truffle Debugger and it will automatically download sources for any verified contracts and allow you to step through them!  Contracts can be verified on either Etherscan or Sourcify.  You no longer need to download these yourself and add them to your Truffle project!  You can see [this](add URL here) earlier post for more information.
```
haltman-at, line 56: The actual URL to add here, since nobody added it, is https://www.trufflesuite.com/blog/debugging-verified-external-contracts-with-truffle-debugger

src/blog/features-from-version-5.md
```diff
@@ -0,0 +1,110 @@
...
+Improved import handling in Vyper
+Here’s another one for our Vyper users; you can now do imports from other projects (via NPM or EthPM) just like in Solidity!
+
+For instance, suppose `examplepackage` is an NPM package which is structured as a Truffle project, and you want to import a Vyper contract `VyperContract` from it.  You can now do
+
+```
```
haltman-at, line 62: Oh, I forgot to stick languages on my triple-backticks.  This one should be Vyper, although that probably won't work, so I guess use Python instead.

src/blog/features-from-version-5.md
```diff
@@ -0,0 +1,110 @@
...
+from examplepackage.contracts import VyperContract
+```
+
+or
+
+```
```
haltman-at, line 68: Also Vyper/Python

src/blog/features-from-version-5.md
```diff
@@ -0,0 +1,110 @@
...
+import examplepackage.contracts.VyperContract as VyperContract
+```
+
+to import it, much like you could do in Solidity!
+
+Note that if your own project uses such “absolute” imports as a way to import its own files from the project root -- e.g., if you have a file `contracts/subdirectory/Contract1.vy` which imports `contracts/Contract2.vy` via
+
+```
```
haltman-at, line 76: Also Vyper/Python

src/blog/features-from-version-5.md
```diff
@@ -0,0 +1,110 @@
...
+import Contract2 as Contract2
+```
+
+then this will work within your own project, but won’t work if other people try to import your `Contract1`.  We suggest that if you want other people to import contracts from your project, you use explicitly relative imports instead:
+
+```
```
haltman-at, line 82: Also Vyper/Python

src/blog/features-from-version-5.md
```diff
@@ -0,0 +1,110 @@
...
+from .. import Contract2
+```
+
+But such imports are now possible.  We hope you’ll find all sorts of uses for this!
+
+### Enumeration values in Truffle Contract objects
+
+When sending a transaction that takes an `enum`, it’s inconvenient to have to look up the numeric constants that the `enum` uses.  Well, now, each Truffle Contract constructor object contains enumeration constants, so you don’t need to do that.  Just do
+
+```
```
haltman-at, line 92: Javascript

haltman-at (changes requested):
URL has to be fixed, the rest is just suggestions.

haltman-at (approved):
Seems good to me!

src/blog/features-from-version-5.md
```diff
@@ -0,0 +1,110 @@
...
+
+In the v5.2-example-box, consider the `contracts/MetaCoin.sol` and `contracts/Conversion.sol`. We would like to import Conversion into the MetaCoin contract to use it but cannot since they require different versions of the Solidity compiler.  To work around this, we have created `contracts/IConversion.abi.json` which contains the ABI from the compiled Conversion. In this way we can now import the interface into `contracts/MetaCoin.sol` (see line 4 in the MetaCoin contract) to use it!
+
+### Solidity stacktraces in Truffle Test
+
+Try running your tests with the `--stacktrace` option (or `-t`) for short and get combined Solidity-Javascript stacktraces!  This feature does have some limitations and is still somewhat experimental, but we expect you’ll find it quite useful.  In the future we may add support for Vyper stacktraces as well!
```
eggplantzzz, line 44: probably should be "option (or `-t` for short)"

src/blog/features-from-version-5.md
```diff
@@ -0,0 +1,110 @@
...
+### Experimental Vyper debugging
+
+You can now use Truffle Debugger with Vyper!  Support for this is still in its early stages; you won’t be able to inspect variables, I’m afraid.  But you can step through a Vyper transaction, and there will likely be more in the future!
+
+But perhaps the biggest improvement to the debugger is...
+
+### Debugging verified external contracts
+
+Use the `--fetch-external` (or `-x` for short) option with Truffle Debugger and it will automatically download sources for any verified contracts and allow you to step through them!  Contracts can be verified on either Etherscan or Sourcify.  You no longer need to download these yourself and add them to your Truffle project!  You can see [this](https://www.trufflesuite.com/blog/debugging-verified-external-contracts-with-truffle-debugger) earlier post for more information.
+Improved import handling in Vyper
```
eggplantzzz, line 57: this should be a heading I think

src/blog/features-from-version-5.md
```diff
@@ -0,0 +1,110 @@
...
+Here’s another one for our Vyper users; you can now do imports from other projects (via NPM or EthPM) just like in Solidity!
+
+For instance, suppose `examplepackage` is an NPM package which is structured as a Truffle project, and you want to import a Vyper contract `VyperContract` from it.  You can now do
```
eggplantzzz, line 60: do we need a couple of ","s here around `VyperContract`?

src/blog/features-from-version-5.md
```diff
@@ -0,0 +1,110 @@
...
+```
+
+But such imports are now possible.  We hope you’ll find all sorts of uses for this!
+
+### Enumeration values in Truffle Contract objects
+
+When sending a transaction that takes an `enum`, it’s inconvenient to have to look up the numeric constants that the `enum` uses.  Well, now, each Truffle Contract constructor object contains enumeration constants, so you don’t need to do that.  Just do
+
+```javascript
+contractInstance.exampleMethod(Contract.ExampleEnum.EnumValue)
```
eggplantzzz, line 93: small thing that doesn't matter much but should we put ";" at the end of the line to make it consistent with the JS above? or we could remove the ";" above as well?

src/blog/features-from-version-5.md
```diff
@@ -0,0 +1,110 @@
...
+    }
+  },
+  // … the rest of your config goes here
+};
+```
+Now you can run `truffle compile` and your project will compile!
+
+In `v5.2-example-box` referenced above, you can see how the `truffle-config.js` is set up to use this new experimental feature. Notice that `contracts/MetaCoin.sol` requires `^0.7.0` and how `contracts/Conversion.sol` requires `^0.6.0`. When running `truffle compile` on this example project you should get a printout of all the versions of the Solidity compiler used during compilation. Namely, you should see that both version `0.6.12` and `0.7.6` are listed.
+
+As mentioned above, when you use this feature, Truffle will analyze each or your Solidity sources one by one to find the appropriate version of the Solidity compiler for each source. Do note, however, that each source file and all of its imports must be compiled with one version of Solidity; this means there must be a version of the compiler that satisfies all of their pragma expressions.
```
eggplantzzz, line 30: Should be "each of your Solidity sources"

src/blog/features-from-version-5.md
haltman-at: Oops, good catch!

src/blog/features-from-version-5.md
haltman-at: Yikes, I missed that, will fix!

src/blog/features-from-version-5.md
haltman-at: Well, the `module.exports = ...` is meant to be a statement; I just meant this to be an expression.  But I'll add a `;` for consistency.

src/blog/features-from-version-5.md
```diff
@@ -0,0 +1,110 @@
+![New Features in Truffle 5.1 and 5.2 Banner](/img/blog/features-from-version-5/blog-header.png)
+
+Truffle 5.1 and 5.2 are exciting releases in the flexibilty they provide to your workflow. Truffle now offers more flexible compilation options by detecting the version of Solidity in your contracts at compile time and by allowing for compilation from ABI files. Because of this, wrangling a project whose contracts use multiple version of Solidity is much easier going forward. We've also enhanced our testing capapbilities with Solidity strack traces and better event decoding, making it much easier to get to the source of an errors and get richer information from events. When our projects integrate with existing protocols, it's a pain to have to deal with those external contracts. Now Truffle makes this easy with the ability to fetch external contracts verified on Etherscan! Vyper support continues to improve with experimental debugging for Vyper contracts, as well as improved import handling for Vyper.
```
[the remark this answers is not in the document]
DiscRiskandBisque: Fixed in commit `79660f9`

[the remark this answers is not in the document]
DiscRiskandBisque: Agreed! Fixed in commit `79660f9`.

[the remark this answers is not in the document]
DiscRiskandBisque: I added a warning before the body content, after the Truffle box mention that calls out three of the features via an alert as experimental. Are there others we need to include?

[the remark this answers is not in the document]
eggplantzzz: nope I think this is fine!

src/blog/features-from-version-5.md
```diff
@@ -0,0 +1,112 @@
+![New Features in Truffle 5.1 and 5.2 Banner](/img/blog/features-from-version-5/blog-header.png)
+
+Truffle 5.1 and 5.2 are exciting releases in the flexibilty they provide to your workflow. Truffle now offers more flexible compilation options by detecting the version of Solidity in your contracts at compile time and by allowing for compilation from ABI files. Because of this, wrangling a project whose contracts use multiple version of Solidity is much easier going forward. We've also enhanced our testing capabilities with Solidity strack traces and better event decoding, making it much easier to get to the source of an error and get richer information from events. When our projects integrate with existing protocols, it's a pain to have to deal with those external contracts. Now Truffle makes this easy with the ability to fetch external contracts verified on Etherscan! Vyper support continues to improve with experimental debugging for Vyper contracts, as well as improved import handling for Vyper.
```
eggplantzzz, line 3: Is this first sentence kind of awkward? Would it be better to write something like "...are exciting releases in that there are some new features to help provide more flexibility to your workflow"?

src/blog/data.json
```diff
@@ -1,4 +1,12 @@
 {
+  "features-from-version-5": {
+    "title": "Features from version 5",
+    "date": "???",
```
eggplantzzz, line 4: We should remember to add the date.

src/blog/data.json
```diff
@@ -1,4 +1,12 @@
...
+    "author": "Harry Altman, Josh Quintal, Tyler Feickert",
+    "published": true,
+    "excerpt": "In this blogpost we will go over some of the features included in version 5.2.x of Truffle as well as some features that you may have missed in 5.1.x.",
```
eggplantzzz, line 7: We should replace this sentence with a better excerpt.

PR #943, opened 2021-03-31
Add documentation for "truffle preserve"

rkalis:
- Add separate page to "Getting Started"
- Add note about "preserve" plugins as opposed to "run" plugins
- Add section about truffle-config environments
- Add section about the "preserve" command to reference
- Slightly expand Filecoin docs section


gnidan (approved):
Nicely written!

PR #955, opened 2021-04-14
Update running-migrations.md

fainashalts:
Adds a section to the Truffle migration documentation to explicitly state that migrations can be done using `async/await` syntax.

### DIFF
--- a/src/docs/truffle/getting-started/running-migrations.md
+++ b/src/docs/truffle/getting-started/running-migrations.md
@@ -254,3 +254,18 @@ deployer.then(function() {
   return b.setA(a.address);
 });
 ```
+
+### Migrations with async/await
+
+You can also migrate your contracts using `async/await`:
+
+Example:
+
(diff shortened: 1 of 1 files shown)


src/docs/truffle/getting-started/running-migrations.md
```diff
@@ -254,3 +254,18 @@ deployer.then(function() {
...
+You can also migrate your contracts using `async/await`:
+
+Example:
+
+```javascript
+module.exports = async function(deployer) {
+  // deploy a contract
+  await deployer.deploy(MyContract);
+  //access information about your deployed contract instance
+  const MyDeployedContract = MyContract.deployed();
```
gnidan, line 269:
```suggestion
  const MyDeployedContract = await MyContract.deployed();
```

gnidan (changes requested):

gnidan (approved):

PR #957, opened 2021-04-18
docs: update pet-shop-box path to be correct

hcliff:
does what it says on the tin 👍


DiscRiskandBisque:
Thanks for this @hcliff! Does it fail to unbox on your machine with the `-box` suffix? If so, that's an error within Truffle. For Truffle's official boxes, we account for the lack of `-box` for a cleaner command.

gnidan:
@hcliff after some team discussion (and since Truffle handles the `-box` suffix automatically), we're going to close this PR to favor the approach of omitting the unnecessary suffix.

Thanks for opening this - would love to hear your thoughts if you have a counterargument!

PR #984, opened 2021-05-26
Update tutorial.md Code block

lukecurtis93:
Minor tweak as the code supplied has an unused var that I believe should be renamed.


fainashalts (approved):
LGTM, thanks!

PR #989, opened 2021-06-01
Tutorial : Updated the reference to semicolon

sethunath:
Updated the reference to semicolon in solidity to use Java as an example instead of javascript, as javascript doesn't require a semicolon to end a statement


fainashalts:
Hi @sethunath, thanks for starting this discussion!  While Javascript doesn't technically require a semicolon, it is common style to use semicolons. We expect our users are familiar with Javascript so we'd prefer to leave this reference.

PR #1006, opened 2021-06-14
Document debugger (y) and (Y) commands

haltman-at:
Realized we didn't have documentation for this, added some.  Included a note about the stopgap nature of their current workings.


cds-amal (approved):
Looks good to me!

gnidan (commented):
It may be worth an additional comment that suggests results may vary, since we're at the mercy of Solidity's sourcemaps

haltman-at:
Mm, that's always true, so I think I'll skip it.

PR #1011, opened 2021-06-19
'balance' should be declared with let

nuel-ikwuoma:
re-assigning a const variable causes an error, hence the needed change.

### DIFF
--- a/src/docs/truffle/testing/writing-tests-in-javascript.md
+++ b/src/docs/truffle/testing/writing-tests-in-javascript.md
@@ -170,13 +170,14 @@ contract("2nd MetaCoin test", async accounts => {
     // Get initial balances of first and second account.
     const account_one = accounts[0];
     const account_two = accounts[1];
+    let balance;
 
     const amount = 10;
 
     const instance = await MetaCoin.deployed();
     const meta = instance;
 
-    const balance = await meta.getBalance.call(account_one);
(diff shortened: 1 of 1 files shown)


cds-amal (approved):
Thanks for the contribution @nuel-ikwuoma!

src/docs/truffle/testing/writing-tests-in-javascript.md
```diff
@@ -170,13 +170,14 @@ contract("2nd MetaCoin test", async accounts => {
     // Get initial balances of first and second account.
     const account_one = accounts[0];
     const account_two = accounts[1];
+    let balance;
```
eggplantzzz, line 173: I would vote that this `let` just gets put on line 180 with the assignment rather than putting it separately up here.

nuel-ikwuoma: 'balance' was used more than once, hence initializing it separately is quite clean.

eggplantzzz: Hmm, I still don't really like it but won't hold it back for that!

nuel-ikwuoma: My pleasure, hopefully more collaborations to come.

haltman-at:
Merging despite yellow because our testing is currently broken.

PR #1029, opened 2021-06-29
Remove claim that calls work with in-test debugging

haltman-at:
This doesn't actually work, but the claim that it does has been up on the website for quite some time; I replaced it with a description of how to work around the problem instead.


gnidan (approved):

haltman-at:
Merging despite yellow because our testing is currently broken.

PR #1034, opened 2021-07-09
Update the docs on testing and the default test network

eggplantzzz:
Part of this work comes from https://github.com/trufflesuite/trufflesuite.com/pull/198.


src/docs/truffle/testing/testing-your-contracts.md
```diff
@@ -25,7 +25,11 @@ To run all tests, simply run:
 $ truffle test
 ```
 
-Alternatively, you can specify a path to a specific file you want to run, e.g.,
+Note that the `test` command will automatically start a test network locally if you have don't have
+a network named "test" defined in the [networks](/docs/truffle/reference/configuration#networks)
```
haltman-at, line 29: You can also specify a network with the `--network` option, and the `development` network also has special behavior.  Maybe rewrite this to:

Note that if you don't specify a network with `--network`, and you don't have a network named "test" or "development" defined in... (etc)

haltman-at: (This isn't a complete description of the behavior, which is a bit messy, but this maybe isn't the place for that.)

eggplantzzz: Ok sounds good, I'll update this.

cds-amal (approved):
I likes.

PR #1039, opened 2021-07-18
Update to ES6 and MetaMask new API

jochasinga:
- Update MetaMask deprecated API: i.e. call to `enable()` is replaced with `request()`.
- Update to ES6: It is safe to say ES6 `async - await` is a new standard all Javascript users know, and it is way more readable than callbacks and promise chaining.
- Minor style refactors: removing trailing spaces and unnecessary code block indentations

NOTE: Either the corresponding functions in the `App` object in the other petshop template should be marked `async` or add instructions here for the user to add themselves for this change.

### DIFF
--- a/src/tutorial.md
+++ b/src/tutorial.md
@@ -286,7 +286,7 @@ You've now written your first smart contract and deployed it to a locally runnin
 
 ## Testing the smart contract using Solidity
 
-<details> 
+<details>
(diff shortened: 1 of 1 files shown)


eggplantzzz:
Thanks for the PR @jochasinga! We'll get it reviewed.

PR #1047, opened 2021-08-19
Add 'pragma' setting to the configuration docs

eggplantzzz:


src/docs/truffle/reference/configuration.md
```diff
@@ -315,10 +315,11 @@ In the `compilers` object you can specify settings related to the compilers used
...
 
 You may specify...
-+ any solc-js version listed at [solc-bin](http://solc-bin.ethereum.org/bin/list.json). Specify the one you want and Truffle will get it for you.
-+ a natively compiled solc binary (you'll need to install this yourself, links to help below).
-+ a dockerized solc from one of images published [here](https://hub.docker.com/r/ethereum/solc/tags/).
++ any solc-js version (using semver) listed at [solc-bin](http://solc-bin.ethereum.org/bin/list.json). Specify the one you want and Truffle will get it for you.
++ "native" to use a natively compiled solc binary (you'll need to install this yourself, links to help below).
++ a dockerized solc tag from one of images published [here](https://hub.docker.com/r/ethereum/solc/tags/).
 + a path to a locally available solc
++ "pragma" to have Truffle autodetect solc versions from your source files. This can be used to compile using multiple versions of solc.
```
cds-amal, line 322: This is not clear to me, perhaps we can offer a use case . A starting point...

> Use this option to resolve the appropriate compiler version for projects referencing modules with different solidity pragmas...

cds-amal (commented):
Just a nit on messaging

src/docs/truffle/reference/configuration.md
eggplantzzz: There is probably another spot where it is more appropriate to go into more detail. This is not the place for what you are describing. Take a look at the descriptions for the other settings; they are all very brief.

eggplantzzz: If we do that, it should probably be a separate PR.

cds-amal: Is there a requirement to keep this section brief? I like the idea of a surfacing more information wherever it goes :)

cds-amal (approved):
Yea, let's get this in, and follow up with another section

src/docs/truffle/reference/configuration.md
```diff
@@ -315,10 +315,11 @@ In the `compilers` object you can specify settings related to the compilers used
 Solidity compiler settings. Supports optimizer settings for `solc`, as well as other settings such as debug and metadata settings.
 
 You may specify...
-+ any solc-js version listed at [solc-bin](http://solc-bin.ethereum.org/bin/list.json). Specify the one you want and Truffle will get it for you.
-+ a natively compiled solc binary (you'll need to install this yourself, links to help below).
-+ a dockerized solc from one of images published [here](https://hub.docker.com/r/ethereum/solc/tags/).
++ any solc-js version (using semver) listed at [solc-bin](http://solc-bin.ethereum.org/bin/list.json). Specify the one you want and Truffle will get it for you.
++ "native" to use a natively compiled solc binary (you'll need to install this yourself, links to help below).
++ a dockerized solc tag from one of images published [here](https://hub.docker.com/r/ethereum/solc/tags/).
 + a path to a locally available solc
```
haltman-at, line 321: IMO this should be clarified to state that this means a path to a locally available soljson, not a solc binary... but also I'm OK with not doing that in this PR and doing that separately later.

src/docs/truffle/reference/configuration.md
```diff
@@ -315,10 +315,11 @@ In the `compilers` object you can specify settings related to the compilers used
 Solidity compiler settings. Supports optimizer settings for `solc`, as well as other settings such as debug and metadata settings.
 
 You may specify...
-+ any solc-js version listed at [solc-bin](http://solc-bin.ethereum.org/bin/list.json). Specify the one you want and Truffle will get it for you.
-+ a natively compiled solc binary (you'll need to install this yourself, links to help below).
-+ a dockerized solc from one of images published [here](https://hub.docker.com/r/ethereum/solc/tags/).
++ any solc-js version (using semver) listed at [solc-bin](http://solc-bin.ethereum.org/bin/list.json). Specify the one you want and Truffle will get it for you.
```
haltman-at, line 318: Just saying "using semver" doesn't make it clear you can input a version range.  This should probably be clarified.  But I'm OK with holding off on that for a separate PR where we just rewrite all this.  So I'm hitting approve on this.

haltman-at (approved):
I have some nitpicks but I think it's fine to merge this now and address them later.

src/docs/truffle/reference/configuration.md
eggplantzzz: No I'm just saying rewriting this section to be something totally different seems like it should be a task different from the one I originally set out to accomplish.

src/docs/truffle/reference/configuration.md
eggplantzzz: agree!

PR #1062, opened 2021-09-07
Changed link at start of drizzle/react tutorial

npham014:
The link in the note at the beginning of the drizzle/react tutorial links to "/tutorials/pet-shop", but it's now hosted at "/tutorial/". Only change is to fix the link.


haltman-at:
Hi, seems like another PR has already made this change!  I'm going to just go ahead and close this, but thank you for taking the time to do this!

PR #1079, opened 2021-10-20
Add options --compile-none and --migrate-none to truffle commands doc

dongmingh:
This PR is to update truffle command test with option --migrate-none which
was supported in truffle test command (see truffle #4340)

### DIFF
--- a/src/docs/truffle/reference/truffle-commands.md
+++ b/src/docs/truffle/reference/truffle-commands.md
@@ -349,7 +349,7 @@ to recognize the plugin. For more information, see [Third-Party Plugin Commands]
 Run JavaScript and Solidity tests.
 
 ```shell
-truffle test [<test_file>] [--compile-all[-debug]] [--network <name>] [--verbose-rpc] [--show-events] [--debug] [--debug-global <identifier>] [--bail] [--stacktrace[-extra]]
+truffle test [<test_file>] [--compile-all[-debug]|--compile-none] [--migrate-none] [--network <name>] [--verbose-rpc] [--show-events] [--debug] [--debug-global <identifier>] [--bail] [--stacktrace[-extra]]
 ```
 
 Runs some or all tests within the `test/` directory as specified. See the section on [Testing your contracts](/docs/testing/testing-your-contracts) for more information.
@@ -361,6 +361,8 @@ Options:
 * `<test_file>`: Name of the test file to be run. Can include path information if the file does not exist in the current directory.
 * `--compile-all`: Compile all contracts instead of intelligently choosing which contracts need to be compiled.
 * `--compile-all-debug`: Like `--compile-all`, but compiles contracts in debug mode for extra information.  Has no effect on Solidity <0.6.3.
+* `--compile-none`: Skip all contract compilations. This option is not intended for use with most tests, which will require compilation.
+* `--migrate-none`: Skip all contract migrations. This option could be useful if you will be deploying your own contracts and don't want Truffle to do this automatically for you.
 * `--network <name>`: Specify the network to use, using artifacts specific to that network. Network name must exist in the configuration.
 * `--verbose-rpc`: Log communication between Truffle and the Ethereum client.
 * `--show-events`: Log all contract events.


src/docs/truffle/reference/truffle-commands.md
```diff
@@ -361,6 +361,8 @@ Options:
 * `<test_file>`: Name of the test file to be run. Can include path information if the file does not exist in the current directory.
 * `--compile-all`: Compile all contracts instead of intelligently choosing which contracts need to be compiled.
 * `--compile-all-debug`: Like `--compile-all`, but compiles contracts in debug mode for extra information.  Has no effect on Solidity <0.6.3.
+* `--compile-none`: Skip all contract compilations.
```
haltman-at, line 364: For these two options, it might be worth adding some prose about when it is appropriate to do this and any dangerous side effects this might have.  (@eggplantzzz, can you say anything about this?)

eggplantzzz: Oof, this is kind of a tough one. `--compile-none` is really kind of like a "you really shouldn't be doing this unless you know what you are doing"-type thing.

`--migrate-none` you could say something like, "this option could be useful if you will be deploying your own contracts and don't want Truffle to do this automatically for you".

src/docs/truffle/reference/truffle-commands.md
```diff
@@ -349,7 +349,7 @@ to recognize the plugin. For more information, see [Third-Party Plugin Commands]
 Run JavaScript and Solidity tests.
 
 ```shell
-truffle test [<test_file>] [--compile-all[-debug]] [--network <name>] [--verbose-rpc] [--show-events] [--debug] [--debug-global <identifier>] [--bail] [--stacktrace[-extra]]
+truffle test [<test_file>] [--compile-all[-debug]] [--compile-none] [--migrate-none] [--network <name>] [--verbose-rpc] [--show-events] [--debug] [--debug-global <identifier>] [--bail] [--stacktrace[-extra]]
```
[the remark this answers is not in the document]
eggplantzzz: So @haltman-at, should it look something like `[--compile-all[-debug]|--compile-none]`?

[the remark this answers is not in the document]
haltman-at: Yeah, something like that?

src/docs/truffle/reference/truffle-commands.md
haltman-at: OK, well, for `--compile-none` we could at least put a warning saying, like, "This option is not intended for use with most tests, which will require compilation.", or something like that.

src/docs/truffle/reference/truffle-commands.md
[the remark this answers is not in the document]
dongmingh: change it, thx

src/docs/truffle/reference/truffle-commands.md
dongmingh: update the doc according to the comments, thx

src/docs/truffle/reference/truffle-commands.md
```diff
@@ -361,6 +361,8 @@ Options:
 * `<test_file>`: Name of the test file to be run. Can include path information if the file does not exist in the current directory.
 * `--compile-all`: Compile all contracts instead of intelligently choosing which contracts need to be compiled.
 * `--compile-all-debug`: Like `--compile-all`, but compiles contracts in debug mode for extra information.  Has no effect on Solidity <0.6.3.
+* `--compile-none`: Skip all contract compilations, this option is not intended for use with most tests, which will require compilation.
```
haltman-at, line 364: This sentence is a comma splice.  It should be split into two sentences.  (Or you could use a semicolon, but I think this should probably be split in two.)

src/docs/truffle/reference/truffle-commands.md
```diff
@@ -361,6 +361,8 @@ Options:
...
+* `--migrate-none`: Skip all contract migrations, this option could be useful if you will be deploying your own contracts and don't want Truffle to do this automatically for you.
```
haltman-at, line 365: Another comma splice; I would again suggest splitting this in two.

haltman-at (commented):
Seems good, just a quick grammar nitpick.

src/docs/truffle/reference/truffle-commands.md
dongmingh: Split into two, thx

src/docs/truffle/reference/truffle-commands.md
dongmingh: split into two, thx

eggplantzzz (approved):
I think this is probably good now.

haltman-at (approved):
Agreed, looks good to me!  @dongmingh, as you'll recall, tests are broken on this repo at the moment, so you'll just have to override that when you merge...

PR #1082, opened 2021-10-31
Update tutorial.md

MislavJuric:
Since `testUserCanAdoptPet()` uses the `expectedPetId` variable, the functions used for testing should be declared after `expectedPetId`, not after the declaration of `Adoption`.


cds-amal (approved):
Thank you @MislavJuric .

haltman-at (approved):
Note that functions actually do not need to be declared after the state variables they use, even if it is conventional.  But it is conventional, so we may as well merge this.

PR #1101, opened 2021-12-07
Add the list subcommand to the config command

eggplantzzz:
Recently a `list` subcommand was added to Truffle in https://github.com/trufflesuite/truffle/pull/4443. This PR documents it. It also adds `build` and `node_modules` directories to the `.gitignore`.

### DIFF
--- a/src/docs/truffle/reference/truffle-commands.md
+++ b/src/docs/truffle/reference/truffle-commands.md
@@ -55,14 +55,15 @@ Options:
 Displays and sets user-level configuration options.
 
 ```shell
-truffle config [--enable-analytics|--disable-analytics] [[<get|set> <key>] [<value-for-set>]]
+truffle config [--enable-analytics|--disable-analytics] [[<get|set> <key>] [<list>] [<value-for-set>]]
 ```
 
 Options:
 
-- `--enable-analytics|--disable-analytics`: Enable or disable analytics.
-- `get`: Get a Truffle configuration option value.
-- `set`: Set a Truffle configuration option value.
+* `--enable-analytics|--disable-analytics`: Enable or disable analytics.
+* `get`: Get a Truffle configuration option value.
+* `set`: Set a Truffle configuration option value.
+* `list`: List all Truffle configuration option values.
 
 ### console
 


kevinbluer (commented):
hey @eggplantzzz, `build` is now replaced by `site` as the build output dir. also given it's now `mkdocs` (python) based, `node_modules` no longer needs to be ignored

eggplantzzz:
Ah ok! I'll remove those changes! I forgot that node_modules were no longer going to be around!

kevinbluer (commented):
Super minor nitpick re the ordering / convention on the shell example, just wondering if something like `[[get <key>]|[set <key> <value-for-set>]|[list]]` might be a little cleaner?

kevinbluer (commented):
Or even `[get <key>|set <key> <value-for-set>|list]`?

cds-amal:
sub commands are hard. Maybe we can take inspiration from `git` for displaying them

src/docs/truffle/reference/truffle-commands.md
```diff
@@ -58,14 +58,15 @@ Options:
...
-truffle config [--enable-analytics|--disable-analytics] [[<get|set> <key>] [<value-for-set>]]
+truffle config [--enable-analytics|--disable-analytics] [[<get|set> <key>] [<list>] [<value-for-set>]]
 ```
 
 Options:
 
 * `--enable-analytics|--disable-analytics`: Enable or disable analytics.
 * `get`: Get a Truffle configuration option value.
 * `set`: Set a Truffle configuration option value.
+* `list`: List all Truffle configuration option values.
```
cds-amal, line 69: Technically these are options but a better term would be sub-command, or config-command?

eggplantzzz: What exactly is the change you are proposing?

eggplantzzz:
@kevinbluer hmm, in Truffle we already have it in the help as 
```
 Usage:        truffle config [--enable-analytics|--disable-analytics] [<list>] [[<get|set> <key>] [<value-for-set>]] 
 ```

It would be nice to keep them both the same, do you think it is worth changing both?

src/docs/truffle/reference/truffle-commands.md
cds-amal: set `Options` to `sub-command`

eggplantzzz: I'd be fine with that, at some point we will then have commands with both an `Options:` and `Sub-commands:` section. Good?

kevinbluer:
Hey @eggplantzzz, apologies for the delay in follow-up to this one...yeah totally agree on consistency. Perhaps we push this one for now (given we want to ensure the parity between `truffle` and the docs) and we discuss any convention changes offline?

eggplantzzz:
Does "push" mean merge it? It is probably more important that the info is documented for now - in other words, this is better than what we currently have, eh?

kevinbluer:
Yep probably should have said "merge" rather than "push" 😀

eggplantzzz:
Ok great, I rebased and I guess this is ready for a final review. Thanks!

kevinbluer (approved):
Alrighty, LGTM

PR #1110, opened 2021-12-14
Refactor Build steps

cds-amal:
- Consolidate building box meta to define_env so it is available in other
mkdocs lifecycles. This removes the need for on_post_build lifecycle
- add ability to limit the local build to 6 boxes to save on network
  lookups. Use as follows:
  ```
  LOCAL_BUILD=true mkdocs serve [-a 0.0.0.0:8000]
  ```

### DIFF
--- a/main.py
+++ b/main.py
@@ -13,24 +13,63 @@ from jinja2 import Template
 def define_env(env):
     "Definition of the module"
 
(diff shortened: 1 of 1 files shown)


kevinbluer (approved):
LGTM

PR #1111, opened 2021-12-14
Update installation and exectution details

cds-amal:
- simplify installation
- update command to run locally


kevinbluer (approved):
Nice one @cds-amal

PR #1115, opened 2021-12-16
broken header fix / temp removing gh stars

kevinbluer:
- header formatting
- temporarily removing stars

### DIFF
--- a/custom_theme/boxes.html
+++ b/custom_theme/boxes.html
@@ -58,7 +58,6 @@
                 </div>
                 <div class="card-body">
                     <h5 class="card-title">{{ box.displayName }}</h5>
-                    <p class="card-text">{{ box.deets.stars }}</p>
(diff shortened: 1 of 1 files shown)


cds-amal (commented):
The boxes header images looks slightly off, do the images need to be updated @kevinbluer ?

kevinbluer:
hey @cds-amal, apologies totally missed this comment at the time...any chance you could provide an example of "slightly off"? given this PR shifts the image to a css background (vs an inline `img`) it might mean some of the images are a bit trimmed

leeftk:
I did my best to try and fix this issue but it seems like the issue is actually with the images for the boxes not being at the specified size in the instructions which is causing the small discrepancies between the different cards. I say let's merge this since its better than having missing images and then we'll as Bryan to change the sizes of the images so they're all the same.

leeftk (approved):
LGTM!!!!!

PR #1127, opened 2022-01-03
fixing nested slug 404 issue

kevinbluer:
Rationale behind fix for reference: "_i think it's because that causes it to generate a `404.html` in the root (with the dynamically generated css files being referenced relatively `assets/css/dynamic.css` vs absolutely `/assets/css/dynamic.css`). as such it's fine for a single slug broken link...e.g. `trufflesuite.com/doh` but not for `trufflesuite.com/doh/doh`. removing the file (really it's just `404.md`) causes mkdocs to handle it differently (e.g. correctly)_"

### DIFF
--- a/src/404.md
+++ b/src/404.md
@@ -1,4 +1,0 @@
----
-template: 404.html
-title: Truffle Suite
----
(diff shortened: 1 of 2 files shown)


kevinweaver (approved):

cds-amal (changes requested):
Hey @kevinbluer,

I'm not sure how to verify the implication of these changes. Can you explain the 404 configuration/generation logic please? Even better, a link to the relevant documentation.

PR #1133, opened 2022-01-05
Added in mkdocs-minify plugin and setup.

michaeljohnbennett:
did some lighthouse testing once implemented.
Seems to be working as expected.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -4,16 +4,19 @@ We welcome all contributions, typo corrections, and general feedback.
 
 ## Contents
 
-* [Branching Processes](#branching-processes)
-* [Merging Processes](#merging-processes)
-* [Build Process & Directory Structure](#build-process--directory-structure)
-* [Layouts](#layouts)
-* [Partials](#partials)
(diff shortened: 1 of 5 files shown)


kevinbluer (changes requested):
noticed that `/dashboard` isn't rendering...it looks like it's because it's in the `custom_theme` directory which doesn't get passed through the markdown renderer and thus leaves `/assets/stylesheets/dashboard.css` and `/assets/js/dashboard.js` as is (e.g. doesn't append `.min` suffix after minification). one option is we just add that manually so it should "just work"...thoughts?

michaeljohnbennett:
ah ok, I will check that out.

michaeljohnbennett:
fixed imports.

kevinbluer:
awesome @michaeljohnbennett, LGTM now! also nice that we're able to take advantage of the fleek build previews (enabled them at the end of last week), e.g. https://bafybeie4nxwwrpx3pyfb37krwt6lo4p6ccvf56mkn5q4sd6mrx2va5r4ay.on.fleek.co/dashboard/

PR #1145, opened 2022-01-14
Fixing all links and some contents

Neurone:
Fixed all the links and some contents, here a more detailed list of modifications.

## Links
- truffleframework.com is gone, but you need to alert the boxes' developers also, because I see them referring truffleframework.com and the old documentation some time
- prefer https over http, where the site supports it
- explicit references to trufflesuite.com are gone
- urls are now relative to root domain, useful to implement automated broken link tests
- few links remain with full domain, and in particular:
  - within source code comments:  (i.e. ./src/docs/tezos/truffle/quickstart.md)
  - with specific config file (i.e. ./src/dashboard/images/favicons/browserconfig.xml)
  - when quoting an old post (i.e. ./src/guides/ethereum-devops-truffle-testrpc-vsts/index.md)
  - when the url is an explicit text and not a link (i.e. ./src/guides/pet-shop/index.md)

Some broken links pointing to missing resources are still present. You can find attached to this PR a summary _grepped_ from [broken-link-checker](https://github.com/stevenvachon/broken-link-checker), where you can check exactly what files are missing, and two full reports from blc and [linkcheck](https://github.com/filiph/linkcheck).

[blc-summary.log](https://github.com/trufflesuite/trufflesuite.com/files/7868150/blc-summary.log)
[blc.log](https://github.com/trufflesuite/trufflesuite.com/files/7867971/blc.log)
[linkcheck.log](https://github.com/trufflesuite/trufflesuite.com/files/7867973/linkcheck.log)

## Contents 
- some miscellaneous fixes
- Truffle Framework became Truffle Suite
- references to Truffle Suite go to /
- references to Truffle go to /truffle

## Note about testing broken links

I tried at first with python based checkers like [linkchecker-mkdocs](https://github.com/byrnereese/linkchecker-mkdocs) and [linkchecker-markdown](https://github.com/scivision/linkchecker-markdown) because they could be included easily in the build process, but with poor results. Then I tried [markdown-link-check](https://github.com/tcort/markdown-link-check), again poor results. I ended up using [broken-link-checker](https://github.com/stevenvachon/broken-link-checker) and [linkcheck](https://github.com/filiph/linkcheck), they gave both very good results even though blc is a lot faster (at least on my rig). Maybe it is not trivial to integrate one of them in the CI in a clear and elegant way, but at least you can easily run them manually after any documentation changes to check for errors.


Neurone:
From my point of view, it's not advisable. 

All internal links are correct with this PR, so there's no need to redirect them. Ideally, it would help run a link checker for every future modification before publishing a new version. This responsibility should be of the dev submitting the PR.

To have the complete documentation fully clean, it remains to find those missing PNGs (or create them again) or remove references to resources not available anymore.

Regarding redirects, even though external sites pointing to old doc pages can take advantage of that technique (explicitly via HTTP 301 or with HTML META), I would not suggest that approach: all can become a mess in a few months because the documentation is updated often, and you should start to keep many versions of the redirects. And sometimes resources are deleted and not available anymore (i.e., info about TruffleCon 2018 and 2019), so you must deal with 404s in any case.

For these reasons, I would prefer an explicit HTTP 404, so both developers and external referrers can correct the situation. Redirects try to avoid this problem, but in my opinion, they are more useful when we are dealing with content linked by applications (i.e., Web APIs) than for documentation.

What do you think?

davidmurdoch:
> All internal links are correct with this PR, so there's no need to redirect them.

SEO is very important to us and external links pointing to these urls will no longer pass page rank to Truffle. The user experience is greatly reduced. Articles and external websites that point to trufflesuite.com links are typically introductory articles for newcomers to the Web3 space; a new user hitting a 404 page when first being introduced to us is not going to leave a good first impression.

If you don't want to to do the work to help rectify the situation we got ourselves into that is perfectly okay; this is still useful work.

Side note: if a page is ever deleted (which I don't think they every should be -- even for old TruffleCon events) the status code _should_ be 410, not 404. Our infrastructure just doesn't afford this status code.

Thanks!

Neurone:
What you say makes a lot of sense; I also agree about not deleting resources even if old, but this is the state of the repo I found, and I don't see a lot of connections with this PR.

This PR fixes internal links pointing to old resources not available anymore because deleted or moved elsewhere. External resources pointing to old links are already receiving 404 responses, with or without this PR.

I suggest creating a specific ad-hoc task/PR to deal with missing links with a companion strategy to keep those redirects updated in the future.

If you think it can be helpful, I can add some of those redirect pages here, but please note that for this work I'm not able to see _all_ the 404s, but just the ones referenced by internal pages. To do exhaustive work, we should check all 404s listed by your web server or Cloudflare.

kevinbluer (approved):
Again, this is awesome...thanks so much for the work on this (it was well overdue).

kevinbluer:
Merging 🥳 Again, as mentioned, totally let us know if / when we can help with any projects you're working on. Thanks again!

PR #1150, opened 2022-01-19
Introducing Ganache 7 blog post

kevinbluer:


src/blog/data.json
```diff
@@ -1,4 +1,11 @@
 {
+  "introducing-ganache-7": {
+    "title": "Introducing Ganache 7",
```
davidmurdoch, line 3: Looks like we still need the "circle" image

davidmurdoch (changes requested):

src/blog/data.json
kevinbluer: Cc @CombsieZ

davidmurdoch: I've added a temporary image so we can ship

davidmurdoch (approved):

cds-amal (approved):
LGTM!

PR #1152, opened 2022-01-24
Adding Google Tag Manager script

kevinbluer:
This PR adds the [GTM scripts](https://support.google.com/tagmanager/answer/6103696?hl=en) to the site.

### DIFF
--- a/custom_theme/main.html
+++ b/custom_theme/main.html
@@ -2,11 +2,23 @@
 
 {% block site_meta %}
     {{ super() }}
-    <!-- <base href="/" /> -->
+    <!-- Google Tag Manager -->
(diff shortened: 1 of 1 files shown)


davidmurdoch:
Reviewing on mobile, so searching is cumbersome... So I may have missed it. Do we currently have a Google Analytics script installed? If so, the tag manager should replace it, as it should then be used to load Google Analytics, AFAIK.

Let's also look into loading it at the bottom of the page so we can prioritize loading content over tracking scripts.

kevinbluer:
Going to go ahead and merge, as Dana suggests "_we should remove it [Google Analytics embed code] but only after successful installation of GTM and subsequent implementation of GA through GTM._"

PR #1175, opened 2022-02-15
Remove extra bit before title in "using the debugger" page

haltman-at:
It's causing it to show up wrong in the sidebar and it was bugging me.


gnidan (approved):
Thanks!

PR #1183, opened 2022-02-24
Document linking to a library at a custom address

haltman-at:
Adds some quick documentation for trufflesuite/truffle#4752 to our `deployer` documentation.


cds-amal (approved):
Thanks @haltman-at. The explanation and example clearly conveys the nuance of the updated behavior.

Issue #1148, opened 2022-01-15
PoC: Using redirects for old urls


Neurone:
Following the discussion with @davidmurdoch and @kevinbluer in PR #1145, I've created a [test branch](https://github.com/InternetOfPeers/trufflesuite.com/tree/redirect-for-old-url) dedicated to the developer's experience of what it means to maintain old URLs with working redirects.

In the dedicated branch, you test the new landing page for the redirects and how to configure the redirect flow with 3 cases:
1. old resource in one of the main folders
2. old resource in a void subfolder
3. old resource in an already populated folder

User's side, I tried with a landing page with 2 seconds of delay, and the UX is not bad.
Developer's side, points 1 and 2 are pretty straightforward; 3 is a bit more annoying because you must explicitly exclude the old pages from the navigation toolbar, adding a dedicated `.pages` file.

I still fear that managing old resources with redirects will become a real pain within a few months. Unfortunately, this is the cleanest and most straightforward method I can think of without touching the webserver's configuration.

gnidan:
Hey @Neurone sorry for the delay; we haven't forgotten about you! The team's just juggling priorities this week and last, but we should have a thoughtful response to your efforts by the end of next week at the latest. Thanks for all your help with these broken link issues!

kevinbluer:
Hi @Neurone, likewise thanks so much for digging into this (and apologies for the delay in following up) 👍

The approach definitely looks viable, albeit with the maintenance considerations you mention. I did do a bit of digging and it does look like there's potentially a more canonical approach via [this plugin](https://github.com/datarobot/mkdocs-redirects) wherein you specify a list of `redirect_maps` in config. Given your prior experience with [mkdocs](https://www.mkdocs.org) just wondered if you might have come across this before and be aware of any pros / cons?

Thanks again!

Neurone:
I didn't know that plugin before, but I looked at it, and I think it's too essential for what you need.

The plugin creates new redirect pages at build time where old ones were, redirecting to the new page. It does it following a list you need to maintain. It does not address problems with old links inside pages or the navigation toolbar, so developers need to do that _and_ manage the new plugin. Also, directory URLs can be a bit tricky to handle with this plugin because it's an on/off feature you apply all over the site.

[...]

Issue #1144, opened 2022-01-13
Found 83 links on trufflesuite.com that 404


davidmurdoch:
See attached report:

[links.txt](https://github.com/trufflesuite/trufflesuite.com/files/7865087/links.txt)

kevinbluer:
Out of interest when did you run this, as this [PR](https://github.com/trufflesuite/trufflesuite.com/pull/1137) got merged ~4 hours ago that fixed a bunch?

davidmurdoch:
I'll run it again now to make sure.

davidmurdoch:
[siteerrors.log](https://github.com/trufflesuite/trufflesuite.com/files/7866227/siteerrors.log)

82 errors found.

Neurone:
They should be all fixed now (#1145), but some resources are actually missing from the repo at all.

kevinbluer:
Thanks again @Neurone! @davidmurdoch this should be all good to close when you have a moment.

lsqproduction:
Fixed in #1145 Thanks again @Neurone for getting to this.

PR #1184, opened 2022-03-01
Fixing legacy guide 404

kevinbluer:


haltman-at (commented):
Hm, is it really a good idea to just move this?  I think we should properly set up a redirect.

kevinbluer:
Can you elaborate on "properly"? Given we're using a statically generated site, we aren't able to leverage web server config to handle redirection / resolution, which gives us a few alternatives (each with trade-offs re maintenance overhead, degree of duplication, etc). Totally could be missing an approach though so would love your take.

- Setup both `src/guides/building-testing-frontend-app-truffle-3.md` and `src/guides/building-testing-frontend-app-truffle-3/index.md` with one containing a redirect to the other. Works but duplicative and a bit of a maintenance pain
- Leverage something like [mkdocs-redirects](https://github.com/datarobot/mkdocs-redirects) and thus move the redirect mapping to config. Cleaner than the above, but potentially overkill if we're just addressing a single link
- Setting a Page Rule / Worker (or equivalent) at the DNS level in Cloudflare. Also potentially overkill and even more of a maintenance pain 🙃 

Also worth noting that the guide in question is from Feb 2017 (referencing Truffle 3) and [technically not published anymore](https://github.com/trufflesuite/trufflesuite.com/blob/main/src/guides/data.json#L83).

haltman-at:
Oh, geez, I sort of assumed we had some sort of redirection solution already set up.  >_>   Like given how often this problem has come up in the past with other pages -- given how many pages we have that have changed URLs over time -- surely we would have a redirection solution set up to handle it, right?  I just sort of assumed that had to have been done by now... by "properly" I just meant "using whatever solution that is to do an actual redirect"!

I mean, I don't really know anything about different ways of setting up redirection and so don't really have much of an opinion between them!  Yeah, the original issue that this was meant to address should probably just be combined into the more general issue for this (obviously this particular page is not especially important as you point out), and this particular PR should probably be closed.  But geez, I had no idea we've been going for this long with the whole redirection problem totally unsolved...

kevinbluer:
Yeah, at least to my knowledge there was never a canonical solution for redirects on the previous version of the site (and such nothing "official" that got carried over to this new iteration). Also afaik things did get moved around at times, without redirects in place, which led to issues like https://github.com/trufflesuite/trufflesuite.com/issues/1144 🤷 

That said, and as per the above thread, it would be great to put something robust in place. There's obviously a few approaches including the suggestion / POC in https://github.com/trufflesuite/trufflesuite.com/issues/1148. However, it would be prudent to ensure that whatever we put in place also accommodates IPFS-based hosting and it's nuance(s). Probably logical for us to discuss in a quick call (will go ahead and schedule something).

PR #1189, opened 2022-03-08
Deprecated the Getting Started with Drizzle and React guide

dkillen:
Unfortunately, Drizzle is currently in need of, possibly substantial, work to update
its dependencies. There are presently too many errors that prevent Drizzle from being
useable. As a result, a user attempting to follow this guide will be unable to complete
those portions that require Drizzle, which is the aim of the guide.

The Getting Started with Drizzle and React guide should be deprecated until such time
as the issues with Drizzle have been resolved. At that time the guide could be
resurrected or a fresh guide prepared.


cds-amal (approved):
Thanks @dkillen.

kevinbluer (commented):
Just pondering an alternate approach to this wherein, rather than deleting, we just toggle the [published state to false](https://github.com/trufflesuite/trufflesuite.com/pull/1189/files#diff-4b02122c1b08bb27b154845355b0735369ca00671b762852954eb87b01d9c9d6L34) so it doesn't show up on `/guides` and put a disclaimer at the top of the page similar to [this one](https://trufflesuite.com/guides/creating-a-cli-with-truffle-3/).

dkillen:
I am not unhappy with that approach. If that is consistent with the way in which we've handled these things in the past, we should probably follow that approach here.

PR #1190, opened 2022-03-09
Removing Teams documentation

kevinbluer:
Removing teams documentation (and associated assets). AFAIK they had already been removed, but it looks like they might have crept back in when we merged the `truffle.dev` repo back into this one where we switch to `mkdocs`, etc.

Worth noting they weren't linked to from anywhere in the site itself, but by virtue of being in `src/docs`, they would have been included in the search index.


cds-amal (approved):
Looks good @kevinbluer!

PR #1197, opened 2022-03-11
describe truffle test exit code

eggplantzzz:
addresses https://github.com/trufflesuite/truffle/issues/2155


tenthirtyone (approved):

cds-amal (approved):
This PR looks good. It reminded me that some[ *nix exit code values are reserved](https://tldp.org/LDP/abs/html/exitcodes.html), is that important?

cds-amal:
This PR clarifies what we do, and should be documented. What we could return, can be an issue we run by the team.

Issue #1104, opened 2021-12-13
Nav bar is broken on desktop in smaller browser widths.


kevinweaver:
## Problem
Currently this happens when you click the hamburger icon in the nav on desktop when their browser is small enough for mobile nav: 
![broken-nav](https://user-images.githubusercontent.com/1020682/145851001-376f8228-1edd-42d2-98f1-c85039fb4df0.gif)

## Solution
Make sure users are can still navigate on desktop in mobile mode.

lsqproduction:
what browser was this? it seems ok on chrome and brave

kevinweaver:
I can reprod this issue in Firefox, Chrome, and Brave.  It looks like "any page other than the home page" might have been too broad.  I am seeing this consistently on Guide pages such as: https://trufflesuite.com/guides/drizzle-and-contract-events/index.html

ameer-khan-ashraf:
@kevinbluer Hi is this issue still open? Just asking as it has been open 2 months ago. I would like to work on this.

kevinbluer:
Hey @ameer-khan-ashraf! 100%, that would be awesome if you're able to work on it. Let me / us know if you have any questions as you go.

Thanks in advance!

ameer-khan-ashraf:
I have found the cause of the issue. In the md files of each post in guides there is a line that hides the material navigation
![image](https://user-images.githubusercontent.com/22253530/160896136-616a9627-dac2-40d3-becb-72d3d188b5b2.png)
Commenting it out makes the navbar work fine. I am guessing @kevinbluer added that to hide navigation for posts in larger screen size? If yes another solution to hide them has to be implemented similar to the one done here.
![image](https://user-images.githubusercontent.com/22253530/160896718-ec2747b6-581c-4d20-8e5a-5854a6ab0ad8.png)

ameer-khan-ashraf:
@kevinbluer I have applied the fix and created a pull request. Please have a look and let me know if that solves the issue.

PR #1212, opened 2022-03-30
Fix for #1104 ( Nav bar is broken on desktop in smaller browser widths.)

ameer-khan-ashraf:
Removes all navigation hide lines from documents in guides, blogs, tutorials, and file box.html.jinja2. 
```
hide:
  - navigation
```
This fixes the issue where the navigation sidebar is not displaying on smaller viewports.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@ site
 src/boxes/*/
 __pycache__
 .DS_Store
+env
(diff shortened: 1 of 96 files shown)


davidmurdoch:
@kevinbluer do you remember why the nav was removed on these pages?

davidmurdoch:
@kevinbluer do you remember why the nav was removed on these pages?

PR #1226, opened 2022-04-14
fix #1217: Exclude irrelevant results from search results

### LINKED ISSUE #1217: Search component includes result not on website
The [website](https://trufflesuite.com/docs/truffle/reference/configuration/) listed `Marymoor park` for `save intermediate` query! It also prioritized this result over possible truffle hits. The search component should constrain the search to truffle's site.  
Screen Shot 2022-04-09 at 7 11 58 AM

xhulz:
The mkdocs-exclude-search plugin was installed (https://github.com/chrieke/mkdocs-exclude-search)

The following folders were excluded from the general search:
      exclude:
        - blog/*
        - events/*
        - staff/*
        - trufflecon2020/*

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -45,6 +45,13 @@ markdown_extensions:
 use_directory_urls: true
 plugins:
   - search
+  - exclude-search:
+      exclude:
+        - blog/*
+        - events/*
+        - staff/*
+        - trufflecon2020/*
+      exclude-unreferenced: true
   - awesome-pages
   - macros:
       module_name: main
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,3 +3,4 @@ mkdocs-awesome-pages-plugin==2.5.1
 mkdocs-macros-plugin==0.6.0
 requests==2.26.0
 mkdocs-minify-plugin==0.5.0
+mkdocs-exclude-search==0.6.4


mkdocs.yml
```diff
@@ -45,6 +45,13 @@ markdown_extensions:
 use_directory_urls: true
 plugins:
   - search
+  - exclude-search:
+      exclude:
+        - blog/*
+        - events/*
+        - staff/*
+        - trufflecon2020/*
+      exclude-unreferenced: true
```
cds-amal, line 54: What does `exclude-unreferenced: true` do?

cds-amal (approved):
Thanks for this @xhulz ! Looks good to me. I left a question in the comment, but that's not reason to hold up this PR.

mkdocs.yml
xhulz: Regarding the docs, files that are added to the docs, but are not listed in the mkdocs.yml page tree: https://github.com/chrieke/mkdocs-exclude-search/issues/17

kevinbluer:
Looks good to me too @xhulz! Only comment I have is why we're completely excluding `blog/*`? From memory it was this entry specifically: `/blog/the-best-things-to-do-in-seattle-during-trufflecon/`, presumably we could just exclude that page only?

xhulz:
Hey @kevinbluer,

In this case, I removed the blog folder because I didn't have any information about what folders we need to keep and what folders we need to remove. But no prob, I will fix that. More suggestions, just let me know. :)

xhulz:
@kevinbluer I've just changed (in the configuration file) to remove only this specific URL. Could you check, please?

kevinbluer (approved):
Awesome, looking good @xhulz 👌

PR #1236, opened 2022-04-22
Bug fix: Replace broken Markdown link with HTML

haltman-at:
With our current setup, Markdown doesn't work inside HTML.  I found a link that didn't work because it was Markdown inside HTML, so I changed it to HTML so it'd work.

Note there might be more of these out there.  I fixed a ton of these in an earlier PR, but it's not like I checked the whole website...


benjamincburns:
Are includes a thing for our website? If so, perhaps it'd be a good idea to break any HTML sections out as HTML files (to be included in to the markdown files), that way we can more easily grep for syntax mismatches?

Or maybe this isn't a super frequent problem and that's just overkill...

benjamincburns (approved):

cds-amal (commented):
Thanks @haltman-at. LGTM

cds-amal (approved):

Issue #686, opened 2020-06-14
writing tests in javascript docs page example #1 is incorrect


blockracers:
https://www.trufflesuite.com/docs/truffle/testing/writing-tests-in-javascript

The first examples should use done or return the promise,
otherwise they it will pass incorrectly eg.

```
it('This will pass but should not', () => {
  Instance.deployed()
     .then(instance => {
          // won't get here
         assert();
      })
});
```
Should be: 

```
it('This will pass but should not',  done => {
  Instance.deployed()
     .then(instance => {
        assert();
         done();
      })
});
```
or
```
it('This will pass but should not',  () => {
  return Instance.deployed()
     .then(instance => {
        assert();
         done();
      })
});
```

DiscRiskandBisque:
Hi @blockracers--thanks for pointing this out and apologies for late response! You're 100% correct, would you be willing to open a PR with these changes? Thanks again!

DiscRiskandBisque:
Hey @blockracers, we're going to convert our tests to use `async`/`await`, so that will supersede this entirely. See issue #772. Keeping this open for now until that's addressed.

PR #1254, opened 2022-06-01
replace .then test example with current async await test

lsqproduction:
This addresses #686.
Removed the `.then()` example.
Updated the `Async await` example with the current code in Metacoin test.


eggplantzzz (approved):
I like it!

eggplantzzz (approved):
I like it!

PR #1264, opened 2022-07-01
Update Ganache docs

MicaiahReid:
This PR removes references to Ganache CLI and adds references to Ganache v7.

This does not remove references to Ganache CLI in any blog posts or in the drizzle and filecoin docs. I am unfamiliar with drizzle and the filecoin Ganache flavor and didn't want to change anything without knowing it was accurate.

Fixes [3152](https://github.com/trufflesuite/ganache/issues/3152)


src/docs/truffle/getting-started/truffle-with-metamask.md
```diff
@@ -118,17 +118,17 @@ Using MetaMask with Truffle Develop is very similar to that of Ganache. The only
 
 In MetaMask, when entering the "New RPC URL", enter `http://127.0.0.1:9545`.
 
-## Using MetaMask with Ganache CLI
+## Using MetaMask with Ganache v7
 
-Using MetaMask with Ganache CLI is also very similar to that of Ganache. The only difference is that Ganache CLI runs by default on `http://127.0.0.1:8545` so you'll want to edit the above web3 code to say:
+Using MetaMask with Ganache v7 is also very similar to that of Ganache. The only difference is that Ganache v7 runs by default on `http://127.0.0.1:8545` so you'll want to edit the above web3 code to say:
```
MicaiahReid, line 123:
```suggestion
Using MetaMask with Ganache v7 is also very similar to that of Ganache UI. The only difference is that Ganache v7 runs by default on `http://127.0.0.1:8545` so you'll want to edit the above web3 code to say:
```

davidmurdoch (changes requested):
I'm not sure I'm comfortable with "Ganache v7" being a proper noun.

tenthirtyone (approved):
Just a q: Would `Ganache` always imply latest version? I understand why you are explicit, e.g. `Ganache UI and Ganache` just doesn't sound right. 

I am comfortable with what you and @davidmurdoch decide about the nomenclature, links/changes look good.

:+1:

Issue #1207, opened 2022-03-29
Cookie policy links to a private google doc


davidmurdoch:
See the Cookies Policy link on https://trufflesuite.com/policy/

PR #1266, opened 2022-07-08
Update cookie policy links

kevinweaver:
Point to the ConsenSys Cookie Policy page instead of linking to a private drive doc. This resolves https://github.com/trufflesuite/trufflesuite.com/issues/1207


kevinweaver:
As a single line, this file is very difficult to review carefully. I can format it to be human readable, otherwise, you can search the text for 4 instances of "cookies policy" to see the new URL.

davidmurdoch (approved):

PR #1291, opened 2022-08-17
fix: redirect has added to tutorial routes

xhulz:
the redirect to `https://trufflesuite.com/guides/pet-shop/` was applied for the routes below:
 - `https://trufflesuite.com/tutorial/`
 - `https://trufflesuite.com/tutorial/ pet-shop/`
 - `https://trufflesuite.com/tutorials/`
 - `https://trufflesuite.com/tutorials/ pet-shop/`


src/tutorial/pet-shop/index.md
```diff
@@ -0,0 +1,773 @@
+---
```
emilyJLin95, line 1: Can you explain how the redirect works?
Why do we need to duplicate the tutorial three times now?

xhulz: `mkdocs redirect` works only pointing an old rote to a new one (1-1). In our case, we have the aggravating factor that every directory is a route (`use_directory_urls = true`), so, we need to physically create all the routes and then put these routes in the routing configuration. In the given case, the petshop was inside the tutorial folder. So we were able to redirect only the tutorial/index.md route to guides/pet-shop/index.md

As new routes appeared (tutorial, tutorials...) we need to implement them physically (create folders and files) and after that, put the 1 to 1 in the yml configuration file.

  - redirects:
      redirect_maps:
        'tutorial/index.md': 'guides/pet-shop/index.md'
        'tutorial/pet-shop/index.md': 'guides/pet-shop/index.md'          
        'tutorials/index.md': 'guides/pet-shop/index.md'
        'tutorials/pet-shop/index.md': 'guides/pet-shop/index.md' 

Unfortunately this is the only way to do the redirect using this plugin. More information here: https://github.com/mkdocs/mkdocs-redirects

Maybe @kevinbluer or @cds-amal might have an idea how to do it another way

kevinbluer:
Hey guys, one approach is to help mitigate duplicate content is just to put a message in the source files that a being redirected from (e.g. `/tutorial/index.md`), such as "Redirecting to /guides/pet-shop". AFAIK this actually gets replaced with the redirect script so there's no chance of it ever being displayed to a user. Also the message also serves as a note to anyone working on the site that the file is a redirect and not the actual source. Thoughts welcome 🙂

kevinbluer:
Also note [8217f5b](https://github.com/trufflesuite/trufflesuite.com/pull/1291/commits/8217f5b8ad9cfe2b5f1789dd90b5db5c7fb4c654) which includes this update.

emilyJLin95 (approved):
LGTM
Can we squash?

PR #1292, opened 2022-08-18
fix: download link

wleven:


eggplantzzz:
Hey @wleven, thanks for the proposed change! However, I don't think this is what we want as the most up-to-date code is in the Ganache repo and not the GanacheUI one.

wleven:
> Hey @wleven, thanks for the proposed change! However, I don't think this is what we want as the most up-to-date code is in the Ganache repo and not the GanacheUI one.

ok

PR #1321, opened 2022-10-03
update boxes page

emilyJLin95:


fainashalts (approved):
LGTM! I was OOO when the bridge box was released, didn't realize it never got updated here. Thanks Emily!

Issue #1288, opened 2022-08-16
Remove the hubspot chat bot on the front page


emilyJLin95:

cliffoo:
[PR](https://github.com/trufflesuite/trufflesuite.com/pull/1325)

It looks like the bot isn't there at the moment due to cors issues.

PR #1325, opened 2022-10-06
Remove hubspot

cliffoo:
Addresses: https://github.com/trufflesuite/trufflesuite.com/issues/1288

So I removed the script that brings in hubspot (which wasn't working btw) but assuming the class names prefixed with `hs-` are hubspot related, I didn't get rid of them because strangely some other parts of the website seem to rely on that styling? I left it in just so nothing else gets messed up.

### DIFF
--- a/custom_theme/partials/footer.html
+++ b/custom_theme/partials/footer.html
@@ -64,11 +64,6 @@
   </div>
 </footer>
 
-<!-- Start of HubSpot Embed Code -->
-<script type="text/javascript" id="hs-script-loader" async defer src="//js.hs-scripts.com/4795067.js"></script>
-<!-- End of HubSpot Embed Code -->
-
-
 <script>
   (function () {
     // handle opening and closing of nav menus


custom_theme/partials/footer.html
```diff
@@ -64,11 +64,6 @@
   </div>
 </footer>
 
-<!-- Start of HubSpot Embed Code -->
-<script type="text/javascript" id="hs-script-loader" async defer src="//js.hs-scripts.com/4795067.js"></script>
```
eggplantzzz, line 68: Do we need to remove the actual JS as well?

eggplantzzz: Oh never mind, I guess this is hosted elsewhere. I'm not familiar with this protocol-less url form here. How does that work? Does it just use the current protocol or something?

eggplantzzz (approved):
noice!

custom_theme/partials/footer.html
cliffoo: Yea it's hosted externally by `hs-scripts.com`. I believe not specifying the scheme means use whatever scheme the current page is loaded in (which makes this a bad idea because it might not be `https?` at all!)

PR #1355, opened 2022-11-15
Update ERC4907 tutorial

emilyJLin95:


kevinbluer (changes requested):
I might be missing it, but in the guide itself I can't see any reference to the point at which you install the OpenZeppelin Contracts lib. The reason I looked is that probably makes sense to explicitly specify a tagged release (e.g. `npm i @openzeppelin/contracts@v4.7.3` if we wanted the previous release before the breaking interface changes in `v4.8.0`). This will help mitigate against any future major/minor breaking changes that would in turn break the steps / example code in the guide.

kevinbluer (approved):
LTGM 👍

PR #1369, opened 2022-11-30
Bold the requirements to avoid potential confusions

ifsheldon:
When I tried truffle for the first time, I find it confusing when I used `truffle migrate` and nothing happened, no warning and no output and no errors. Later I realized that although I wrote the script for deployment, the filename of the script was wrong, not prefixed with a number. I didn't read the documentation carefully and/or it's not obvious enough, so here's my PR lol. If I missed anything, just let me know. Thanks!


cds-amal (approved):
This seems useful

haltman-at (approved):

PR #1378, opened 2022-12-15
snaps guide

emilyJLin95:


src/guides/metamask-snap/index.md
```diff
@@ -0,0 +1,1318 @@
...
+if(matchingFunction.length > 0) { 
+  returnObject.method = matchingFunction; 
+}
+```
+
+After you reconnect your snap, you should see something like this:
+
+![function insight](./img/function_insight.gif)
+
+To get even more granular, let's actually decode the argument data when `withdrawNFT` is called. First, add this function, which will convert our data to strings:
```
Montoya, line 1099: This function converts certain arguments which are not JSON serializable to strings. I don't know if that applies to all of our data.

src/guides/metamask-snap/index.md
```diff
@@ -0,0 +1,1318 @@
...
+    }
+    returnObject.readResult = ethersReadResult; 
+
+  } catch (err) {
+    returnObject.canWithdraw = `${err}`;
+  }
+}
+```
+
+Now, when you call `withdrawNFT`, the snap should give you some information about whether or not the NFT can be drawn and who the owner and second signer are!
```
Montoya, line 1167: can be **with**drawn

Montoya: Also worthwhile to note that this function is reading data from the chain, using the getApproval method we wrote in NFTVault.sol, and we can do that because withdrawNFT has the same arguments that we need for getApproval. So we are getting the arguments used for withdrawNFT and passing those to the onchain contract getApproval method to see if the NFT is approved for withdrawal or not. 

Plus, as a bonus, the Snap could be updated to use this information: 

- To tell the user the address of the second signer
- To tell them if they are attempting to withdraw an NFT they do not own
- To tell them if the NFT is not found (the first two values will be 0x0)

src/unleashed/index.md
```diff
@@ -113,5 +113,16 @@ hide:
...
                 </div>
             </div>
+            <div class="row mt-5">
+                <div class="col-md-3">
+                    <img src="/img/unleashed/W3U-ep-10.png" alt="Web3 Unleashed Episode #10" class="rounded-circle img-fluid green" />
+                </div>
+                <div class="col-md-9">
+                    <h2 class="mt-0 font-weight-bold">Episode #10 - How to build a MetaMask snap</h2>
+                    <p>12/15/2022 9:00 PST&nbsp;&nbsp;&nbsp;<i class="far fa-user"></i> Emily Lin</p>
+                        <p>In this episode, we'll be joined by Christian Montoya and Guillaume Roux to talk about extending the functionality of your MetaMask wallet using MetaMask snaps!</p>
```
Montoya, line 123: using MetaMask **S**naps!

Montoya (changes requested):
Various fixes

emilyJLin95:
@Montoya addressed comments!

kevinbluer:
Heads up it's not currently rendering (see [current build preview](https://bafybeidfi46t4ek5sw7pcoljpq35ev6the2bv3xpg22ktv2e7jmxo6nj6q.on.fleek.co/guides/metamask-snap/)). I think it's due to the use of double braces in the code samples which are being misinterpreted by [mkdocs](https://www.mkdocs.org/).

Screenshot 2022-12-16 at 11 58 30 AM

I have a feeling this has come up before so let me see if I can dig out a workaround.

Montoya:
> @Montoya addressed comments!

Looks good to me, sounds like once the rendering issue is fixed this is good to go.

kevinbluer:
Alrighty, fix in place. Solution was slightly hacky in that it involved putting a space between the double curly braces (during the render phase I believe mkdocs looks out for them and treats them as objects). There only a few instances so it's only a minor stylistic thing and won't effect things for users that copy the code.

Screenshot 2022-12-16 at 4 35 47 PM

If all good with it, good to publish from my side 👌

Montoya (approved):

emilyJLin95:
@kevinbluer thabjs for investigating the issue!

squashed the commits

src/guides/metamask-snap/index.md
```diff
@@ -0,0 +1,1324 @@
...
+### Create `packages/snap/src/insights.ts`
+
+First, we'll need to install the necessary dependencies:
+
+```shell
+cd ../snap
+yarn add @metamask/abi-utils ethers
+```
+
+Then create a new file called `insights.ts`. Add this code, and remember to replace `SimpleNFTContractAddress` and `NFTVaultContractAddress` with your migrated contracts. You can access those quickly by calling `truffle networks` from the `packages/truffle` folder.
```
kevinbluer, line 997: It looks like the need to replace `SimpleNFTContractAddress` and `NFTVaultContractAddress` is now redundant?

src/guides/metamask-snap/index.md
```diff
@@ -0,0 +1,1324 @@
...
+
+Plus, as a bonus, the Snap could be updated to use this information:
+
+1. To tell the user the address of the second signer
+2. To tell them if they are attempting to withdraw an NFT they do not own
+3. To tell them if the NFT is not found (the first two values will be 0x0)
+
+![decoded arguments](./img/decodearguments.gif)
+
+The completed code looks like this:
```
kevinbluer, line 1177: As per the earlier comment it looks like the need to replace the hardcoded contract addresses has been updated in favor of get it dynamically

kevinbluer (approved):
Including a couple of additional minor comments. Also approving so you can merge whenever ready. Awesome tutorial too btw!

PR #1382, opened 2023-01-05
truffle + security draft

kevinbluer:


davidmurdoch (approved):
Approving the fix for the close-sidebar-link, even though it causes a new issue

PR #1391, opened 2023-01-23
add section about overloaded functions to contract reference docs

eggplantzzz:
A new section about overloaded functions was added to @truffle/contract's README with [this PR](https://github.com/trufflesuite/truffle/pull/5836). This should also be reflected by the docs on the website.


cds-amal (approved):
Looks good. we should get this in.

PR #1398, opened 2023-02-09
fix: update some bad links

Hyperzsb:
Signed-off-by: Shaobo ZHANG <hyperzsb@qq.com>


kevinbluer (approved):
Awesome, many thanks for these @Hyperzsb 👍

PR #1431, opened 2023-07-08
Ultimate guide to truffle: Fix development network object for Ganache CLI

d-velev:
Small error: there is no "," after the host property


cliffoo (approved):

eggplantzzz (approved):
Thanks for catching that!

